PR #7, opened 2018-08-01
Start writing tests...

mwcraig:
...right now these mostly fail, will work on in the morning.

There is one problem I'm hitting @pllim -- when I try `python setup.py test` locally I get an error when tests are being collected; any idea what the cause might be? In the meantime I'm running them with `python setup.py test --args '--continue-on-collection-errors'`

```python-traceback
============================================================ ERRORS =============================================================
_____________________________________________ ERROR collecting astrowidgets/core.py _____________________________________________
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/site-packages/pytest_doctestplus/plugin.py:95: in collect
    for test in finder.find(module):
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/site-packages/pytest_doctestplus/plugin.py:340: in find
    extraglobs)
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/doctest.py:933: in find
    self._find(tests, obj, name, module, source_lines, globs, {})
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/doctest.py:992: in _find
    if ((inspect.isroutine(inspect.unwrap(val))
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/inspect.py:512: in unwrap
    while _is_wrapper(func):
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/inspect.py:503: in _is_wrapper
    return hasattr(f, '__wrapped__')
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/site-packages/ginga/misc/Bunch.py:211: in __getattr__
    return self.tbl[attr]
/Users/mcraig/conda-main/envs/awid-dev/lib/python3.6/site-packages/ginga/misc/Bunch.py:96: in __getitem__
    return self.dict[k][1]
E   KeyError: '__wrapped__'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
==================================================== 1 error in 0.43 seconds ====================================================
```


pllim:
@mwcraig , do we need `doctestplus` on `ImageWidget` tests? It appears incompatible with Ginga's `Bunch` object, which duck-type `dict` but only to some extend.

astrowidgets/tests/test_api.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+from astropy.io import fits
+from astropy.nddata import NDData
+from astropy.table import Table
+
+from ..core import ImageWidget
+
+
+def test_load_fits():
+    image = ImageWidget()
```
pllim: My suggestion is to make these functions into a test class' methods. That way, you can initialize `image = ImageWidget()` only once at `setup_class()` and then use the same object throughout.

mwcraig:
> @mwcraig , do we need doctestplus on ImageWidget tests? It appears incompatible with Ginga's Bunch object, which duck-type dict but only to some extend.

No, probably not; I just tried setting `doctest_plus = disabled` but still got the same error, though...

astrowidgets/tests/test_api.py
mwcraig: Sounds reasonable; I'm going to keep implementing as functions for the moment so we can see whether there are any tests where we want separate `ImageWidget` instances. 

For example, it looks like setting `cuts` with a string fails in a way users will find surprising if the data hasn't been set yet. In that case the `cuts` are set to `(0, 0)` (which makes sense) but that is probably not what the user wants; to test for this I think we want a new `ImageWidget` without data for that test.

pllim:
@drdavella , any advice on how to get rid of the `doctest_plus` error? Thanks.

drdavella:
It's possibly due to some regressions introduced in pytest 3.7, but it's hard for me to tell without the full test report.

pllim:
In that case, does the error go away if you use `pytest <= 3.6`, @mwcraig ? Thanks, Dan!

mwcraig:
All of the tests currently pass 😀 so read to merge....

mwcraig:
Unfortunately I'm using pytest 3.6.4 😢 

I'll take another look in the morning...

drdavella:
If you want to post the full log here I can take a look and see if anything obvious stands out.

pllim:
@mwcraig , now I am confused. Is the `pytest` failure still a problem or not?

mwcraig:
If I disable doctest_plus (by removing it from setup.cfg) the tests run successfully, so that's what I'll do later today, I think, unless we find a better fix before that...

pllim:
The only thing that I image `doctest` would try to run is this in the docstring:

```python
            from ginga.misc.log import get_logger
            logger = get_logger('my_viewer', log_stderr=False,
                                log_file='ginga.log', level=40)
```

Even then, I am not sure why it is choking on `ginga.misc.Bunch`, which is not directly mentioned anywhere. :woman_shrugging:

pllim:
@mwcraig , :+1: to disabling doctest for now. Figuring out how to make it work, if needed, can be a separate issue.

pllim:
@mwcraig , I activated https://travis-ci.org/astropy/astrowidgets . FYI

mwcraig:
🎉 triggered the first build on travis! 🎈

pllim:
The Sphinx failure is because `astropy_helpers` you are using is missing patch from astropy/astropy-helpers#368 . I recommend changing the submodule here to point to its current `master` and copy over a compatible `ah_bootstrap.py` file accordingly.

pllim:
Or maybe missing dependencies for Sphinx... 🤔

mwcraig:
Is there any reason *not* to use conda-forge in the tests? That would pick up ipyevents as a conda package...

mwcraig:
I think I fixed the helpers problem....I had added it as `astropy-helpers` not `astropy_helpers`. Second time in a month I've lost time to a `_` vs `-` issue 🙄

pllim:
Recent discussions suggested that conda is slower than pip. Otherwise I don't think any reason it won't work.

pllim:
It passed! :tada:

mwcraig:
Do you want me to squash before merging?

pllim:
I think @eteq and I have differing opinions on the topic of squashing... But I say yes. ;)

mwcraig:
I let GitHub do its squash and merge magic....

PR #14, opened 2018-08-02
add width/height properties

eteq:
Pretty uncontroversial - just adds some width/height properties to check what the widget was created with.

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -696,3 +696,18 @@ class ImageWidget(ipyw.VBox):
         Save out the current image view to given PNG filename.
         """
         self._viewer.save_rgb_image_as_file(filename)
+
+    @property
+    def width(self):
+        """
+        The width of the widget in pixels
+        """
+        return int(self._jup_img.width)
+
+    @property
+    def height(self):
+        """
+        The height of the widget in pixels
+        """
+        return int(self._jup_img.height)
+


eteq:
NOTE: this branch should be deleted when merged/closed - it's only in this repo because it was prior to the migration to `astropy` (and would normally have been in my fork)

mwcraig:
The standard in ipywidgets is to store the width in the layout property; we should be doing that in `__init__` and should have users access the width/height that way. Though `ipywidgets.Image` also has a width/height those don't reliably determine the width/height of the image; that isn't an ipywidgets (mis)feature. The problem is that any css applied to the image will overwrite the width/height attributes in an `` tag. 

I'll open a separate PR in a bit to redo the width/height of the `Image` widget in initialization and add an example of accessing the width/height. And we'll need to hash out what we mean by widht/height on initialization. Quickly, though:

```python
image = ImageWidget()
# To get the width of the Box:
image.layout.width
# To get the height of the Box, which won't be the height of the image if the cursor position is displayed
image.layout.height
```

pllim:
Superseded by #22

pllim:
@eteq , is it ok to delete `add-width-height` branch?

mwcraig:
>  is it ok to delete add-width-height branch?

Hope so, just deleted it 🙀

eteq:
@pllim @mwcraig - oops, it was an accident that I had it here.  Normally I wouldn't want to delete that until #22 is in, but I think this was just an oversight that it was in this fork because I migrated this from `eteq`...

Issue #9, opened 2018-08-02
Clarify zoom API


mwcraig:
Right now `zoom_level` can be get/set or one can call `zoom` which seems to do the same thing. Not sure that was the intent in the original API proposal at https://gist.github.com/eteq/bb625148969db51b2b03b71e92ca5231

eteq:
Oh, it's about relative vs absolute zoom.  That is get/set of `zoom` is all that is needed for the *functionality*, but something along the lines of this is useful:
```
def zoom(factor):
    self.zoom_level = self.zoom_level*factor
```
So it's really just trivial syntactic sugar.  Now that I look at that, perhaps an alternative is to just tell users to do ``self.zoom_level *= factor``, though?  As long as `zoom_level` is a *multiplicative* factor rather than a non-linear thing like some other js image viewers I've seen, anyway.

eteq:
I implemented what I just said above in #15 ... 

*however*, that made me realize a confusing element: `ginga` distinguishes between "zoom" and "scale".  As I understand it, scale is 2d (x and y), while zoom is a fixed change.  But zoom is also distinguished by the "rate" being settable.  Presumably this is to facilitate user interaction to make the zoom be faster.

My instinct is that the API we're using here is a bit better for the purpose we have here because it's simpler, and is what I think more people have in mind when they think of "zooming" - but it's a lot closer to what ginga calls "scale".  And maybe we *do* want a way to control the "zoom rate".  I just think that should be handled as a widget property and not better of the Python-useful API.  Do you have any thoughts on that @pllim ?

pllim:
Looks like #15 is merged. I'm okay with current implementation. We can revisit extra functionality if users complain about not able to set zoom the other way. Should we close this?

mwcraig:
Closed by #15

PR #15, opened 2018-08-02
Zoom updates

eteq:
This implements the idea I was suggesting in #9


eteq:
NOTE: this branch should be deleted when merged/closed - it's only in this repo because it was prior to the migration to `astropy` (and would normally have been in my fork)

PR #22, opened 2018-08-05
Add properties for image width/height

mwcraig:
This is an alternative to #14. 

There are really two sizes here:

+ The width/height of the image
+ The width/height of the ipywidgets Box containing the image and the cursor display. 

I guessing most users will care more about the *image* size; to make explicit that that is what is settable I've changed the names to `image_width` and `image_height`. 

Ideally those would be set to the width/height of the *layout* of the image widget rather than the image itself. The reason is that any css that applies to the image will override the img width/height tags. That means, for example, that setting the width of the `ipywidgets.Vbox` using the `layout` of the `VBox`, which applies the css width to everything in the box, will override the img width/height.

If we set width/height of the layout then it is applied via css. In addition, setting the width/height this way allows more flexible values (e.g. as a percent). 

edit: reference for image width/height on the img attribute vs via CSS: https://www.w3schools.com/html/html_images.asp

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -464,7 +464,8 @@ class ImageWidget(ipyw.VBox):
         xy_col = np.asarray(xy_col)  # [[x0, y0], [x1, y1], ...]
 
         if include_skycoord:
-            radec_col = np.asarray(radec_col)  # [[ra0, dec0], [ra1, dec1], ...]
+            # [[ra0, dec0], [ra1, dec1], ...]
+            radec_col = np.asarray(radec_col)
 
             # Fill in X,Y from RA,DEC
             mask = np.isnan(xy_col[:, 0])  # One bool per row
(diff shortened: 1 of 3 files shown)


mwcraig:
Yeah, probably should. The user can set it right now with `._jup_img.layout.width = ...`

pllim:
> Yeah, probably should. 

So, adding a setter should solve that problem?

pllim:
Matt, do you plan to add the setters here, or should we merge?

mwcraig:
I'll add setters tonight, thanks for the reminder!

mwcraig:
Riiight...so now I remembered why this is a little tricky.  Turns out that setting a width on the layout of the image doesn't actually set width of the image, but a container with a narrower width (e.g. the outer VBox) will override the image's width. 

This is really a little messed up and should be dealt with at the ipywidgets level. Will stew on what to do about it here overnight.

mwcraig:
Sorry for the delay -- I'll add setters this week. This (using `layout.width`) is the best way to go for now.

mwcraig:
I have no clue what is wrong with the flake test...merging despite that fail...

PR #33, opened 2018-08-06
DOC: Setup doc layout, add installation notes

### LINKED ISSUE #18: Installation notes
Somewhere (README? Docs?) there should be a bit more on the installation.  Don't have time to take care of it right now, but some bullet points I'm reminding myself to include because they tripped me up:

For jupyterlab:
* Needs an up-to-date jupyterlab
* Needs a reasonably up-to-date nodejs
* Need to have `ipywidgets` and `ipyevents` *in the kernel of execution*
* Need to do ``jupyter labextension install @jupyter-widgets/jupyterlab-manager`` *in the kernel of jupyterlab*
* Need to do ``jupyter labextension install @jupyter-widgets/jupyterlab-manager ipyevents`` (possible duplicate of the above, but I had to do them in order to get it to build properly...)

(I still haven't gotten ipyevents working though, so might have more to add)

pllim:
Fix #18

### DIFF
--- a/astrowidgets/__init__.py
+++ b/astrowidgets/__init__.py
@@ -12,10 +12,10 @@ import sys
 
 __minimum_python_version__ = "3.5"
 
-class UnsupportedPythonError(Exception):
-    pass
-
 if sys.version_info < tuple((int(val) for val in __minimum_python_version__.split('.'))):
+    class UnsupportedPythonError(Exception):
+        pass
+
     raise UnsupportedPythonError("astrowidgets does not support Python < {}".format(__minimum_python_version__))
 
 if not _ASTROPY_SETUP_:
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ edit_on_github = True
 github_project = astropy/astrowidgets
 # install_requires should be formatted as a comma-separated list, e.g.:
 # install_requires = astropy, scipy, matplotlib
-install_requires = astropy, ipywidgets, ipyevents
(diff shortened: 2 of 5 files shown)


astrowidgets/__init__.py
```diff
@@ -12,10 +12,10 @@
 
 __minimum_python_version__ = "3.5"
 
-class UnsupportedPythonError(Exception):
-    pass
-
 if sys.version_info < tuple((int(val) for val in __minimum_python_version__.split('.'))):
+    class UnsupportedPythonError(Exception):
```
pllim: Note: Need to bury the exception class here to avoid `automodapi` to pick it up.

setup.cfg
```diff
@@ -47,7 +47,7 @@ edit_on_github = True
 github_project = astropy/astrowidgets
 # install_requires should be formatted as a comma-separated list, e.g.:
 # install_requires = astropy, scipy, matplotlib
-install_requires = astropy, ginga, pillow, ipyevents, opencv-python
```
pllim: Note: I had problem installing with `opencv-python` in the requirements even though I have `opencv` installed. So have to remove this one. Not sure how this didn't trip the CIs. :woman_shrugging:

mwcraig: I'll try to explore in the next day or two. Could we keep `ginga` and `pillow`? At the moment they are needed to use the widget at all..

pllim: OK I added them back in `setup.cfg`.

Issue #36, opened 2018-08-22
Setting cursor position no longer works


pllim:
In the example notebook, there is a cell to do `w.cursor = 'top'` to move X/Y/RA/DEC/Value display to the top of the image display. It used to work but no longer does (at least it has no effect when tested on Windows 10). Will have to check to see if it is also broken on `*nix`.

pllim:
So, here is what is happening now... This in `core.py` used to work when the widget class used to be in the notebook itself. But now it does not work anymore.

```python
    def _repr_html_(self):
        """
        Show widget in Jupyter notebook.
        """
        from IPython.display import display
        return display(self._widget)
```

For cursor position to update properly, instead of calling `w` in the notebook, one needs to call `display(self._widget)` instead, which is not very elegant. @mwcraig , any ideas?

PR #41, opened 2018-08-24
Fix cursor setter

mwcraig:
As reported in #36, that never got updated when we switched to subclassing from `VBox` instead of having a custom `_repr_html`.  Can you please confirm that this work for you @plim?


mwcraig:
> p.s. I added a commit to update notebook instructions here. Hope you don't mind.

Thanks!

PR #46, opened 2018-10-29
Fix add_markers

### LINKED ISSUE #45: add_markers modifies the input data table if pixel_coords_offset is not zero

mwcraig:
This fixes #45 (the problem that add_markers modifies the input table).

### DIFF
--- a/astrowidgets/_astropy_init.py
+++ b/astrowidgets/_astropy_init.py
@@ -50,8 +50,8 @@ if not _ASTROPY_SETUP_:  # noqa
                 try:
                     update_default_config(__package__, config_dir)
                 except ConfigurationDefaultMissingError as e:
-                    wmsg = (e.args[0] +
-                            " Cannot install default profile. If you are "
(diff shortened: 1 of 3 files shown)


astrowidgets/_astropy_init.py
```diff
@@ -50,8 +50,8 @@
                 try:
                     update_default_config(__package__, config_dir)
                 except ConfigurationDefaultMissingError as e:
-                    wmsg = (e.args[0] +
-                            " Cannot install default profile. If you are "
+                    wmsg = (e.args[0] + " Cannot"
```
pllim: p.s. I never completely agree with this particular PEP 8 check. I am open to adding stuff to be ignored for `[flake8]` if they become too cumbersome.

mwcraig: this seems fine -- I was really onfused last night for a while because I needed to rebase to see what lines were causing the issue.

PR #48, opened 2018-10-29
Add pixel_coords_offset attribute

### LINKED ISSUE #47: Make pixel offsets more explicit
RIght now in some places you can specify whether you want pixel numbering to start at 0 or 1 (e.g. in [`add_markers`](https://github.com/astropy/astrowidgets/blob/master/astrowidgets/core.py#L496)) but in other places an offset of 1 is baked in, for example in [the method that calculates the pixel location](https://github.com/astropy/astrowidgets/blob/master/astrowidgets/core.py#L164) to display.

I'm thinking that setting that preference once when the object is created might be a better way to go. 

Also in [the other mouse callback](https://github.com/astropy/astrowidgets/blob/master/astrowidgets/core.py#L164).

mwcraig:
This fixes #47 by adding a pixel_coords_offset to the initializer and using it wherever the offset comes up.

I can add a property/setter if desired.

I did sneak in one other change: value is now displayed to three decimal pplaces

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -48,6 +48,12 @@ class ImageWidget(ipyw.VBox):
         e.g., rotation and mosaic. If this is enabled and you
         do not have ``opencv``, you will get a warning.
 
+    pixel_coords_offset : int, optional
+        An offset, typically either 0 or 1, to add/subtract to all
+        pixel values when going to/from the displayed image.
+        *In almost all situations the default value, ``0``, is the
+        correct value to use.*
+
     """
 
     def __init__(self, logger=None, image_width=500, image_height=500,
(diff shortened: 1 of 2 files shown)


mwcraig:
Oh, I should probably point out that the current default value is 0 but we can change that :)

mwcraig:
This is not ready to merge -- found some more pixel offsets...

pllim:
Actually, I am not 100% sure anymore... `origin` is set in the WCS methods like https://github.com/ejeschke/ginga/blob/198ac29add65aaf6f6c44680b8b8a83132b30e6e/ginga/util/wcsmod/wcs_astropy.py#L145 but it is not immediately clear to me how that hooks up to `pixel_coords_offset` that is set in `general.cfg` (https://github.com/ejeschke/ginga/blob/62edc04ca9796b038b67b486ad3adb36e27f703e/ginga/rv/plugins/Cursor.py#L21-L25). @ejeschke, can you please advise? Thanks!

ejeschke:
`pixel_coords_offset` is a setting that you can put in your `$HOME/.ginga/general.cfg`, *but it only affects the reference viewer, and only for the `field-info` callback, which is what updates the `Cursor` plugin (this is a reference viewer only callback). 

Regular mouse position type callbacks with the basic Ginga widget will all report a data position that is 0 based. This is consistent with standard treatment in `astropy` and other Python data indexing schemes.

mwcraig:
Assuming tests pass I'll go ahead and merge this. I've left the ability to set a pixel offset in the package in case someone finds it useful down the road but added the advice to the docstring to always leave it at 0.

mwcraig:
...and made some effort to make it hard to change after the widget has been constructed.

pllim:
Thanks, Matt!

PR #59, opened 2019-01-02
Fix jupyter lab compatibility

### LINKED ISSUE #44: Ginga zoom keys conflict with notebook/lab heading keys
I think this is a jupyterlab thing (i.e., it might not work this way in the notebook), but when I have the widget up and hit a number key to zoom, the event apparently gets passed onto the lab and it will change whatever cell I have selected into a header-level (the default Jupyter behavior).  If it happens to be the cell containing the image widget, it hard-crashes the lab and I have to refresh the browser to get anything to work.

Sort of vaguely related to #39 since it might be the easiest way around it

mwcraig:
The real work has been done in ipyevents; this simply sets a new lower bound for the version of that package.

One side effect of the changes is that the image widget now has a blue border around it when it has focus. I don't think that is a problem, but if we want to address we can by adding the appropriate styling (though I'd rather do that in a separate PR).

This fixes #44 and I think also mwcraig/ipyevents#27

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ edit_on_github = True
 github_project = astropy/astrowidgets
 # install_requires should be formatted as a comma-separated list, e.g.:
 # install_requires = astropy, scipy, matplotlib
-install_requires = astropy, ginga, pillow, ipywidgets, ipyevents
+install_requires = astropy, ginga, pillow, ipywidgets, ipyevents>=0.4.0
 # version should be PEP386 compatible (http://www.python.org/dev/peps/pep-0386)
 version = 0.1.0.dev0
 # Note: you will also need to change this in your package's __init__.py


mwcraig:
Merging, since this is pretty straightforward. @pllim @eteq -- astrowidgets should now work with lab.

pllim:
I am not sure if jupyter lab compat is really fixed or not. Or maybe I just don't know how to focus the widget properly (I see blue line on the left but not blue border all around).

Jupyter lab usage was never broken for me before this patch, but now it is. Using the latest dev of astrowidget, dev Ginga, dev Astropy, and ipyevents 4.0, when I mouse over the displayed image, the X/Y/RA/DEC/value info bar does not update anymore. It updates when I use jupyter notebook though.

Should I open a new issue about this?

mwcraig:
Yes please -- can you also include the jupyterlab version?

Also, maybe try this: `jupyter labextension install ipyevents`

Issue #16, opened 2018-08-02
select_points / session-ful marking


eteq:
This is forked from https://github.com/eteq/astrowidgets/pull/1#issuecomment-406811166 + related discussion in #10.

Specifically, about the `select_points` function from the API - it was replaced by ``w. is_marking = True``.

My argument is that, while having `marking` be settable is tempting at first glance, I think it's better as a method because it's actually *doing something*.  That is, the typical user (I think) isn't thinking about the act of marking stuff as a state machine, but rather a set of actions. Most workflows are either:

workflow 1:

1. show an image
2. mark some points/regions on it that are defined in a table or similar that requires no interaction
3. zoom around/look at what's marked
4. possibly repeat 2-3, or workflow 2

workflow 2:

1. show an image
2. interactively click on some objects in the image
3. do something in code with the marked locations
4. possibly repeat 2-3 for a different set of points, or workflow 1

To be very concrete one might select a subset of stars that are good PSF stars, and then use them to build a PSF (workflow 2), then the user wants to mark all the *new* locations of the stars after re-centroiding/fitting (workflow 1), then they want to mark a different set of stars from that set that turn out to look bad from the fitting (workflow 2 again).

So that sort of use case motivates `select_points`/`mark_points`. It's not modal per se (at least not to the user), but rather they start the action of selecting points, then they finish and are done.  Then the *next* time they want a *different* set of points.

TL;DR: While I am fine with the name changes, I think `select_points`/`mark_points` is still useful, and it *should* retain the behavior of knowing which "session" the selections were made from.  Perhaps `get_markers` returns all of them, but `get_last_markers()` or something like that returns only the most recent set?  Or `get_marker_sessions()` returns a list of lists of markers?

pllim:
If we want to support "sessions", then we also need to support merging/splitting/deleting/saving/loading sessions? Or am I over-thinking this?

mwcraig:
I'm also inclined at this point to switch from making `self.is_marking` a settable property to adding a method called `start_marking` for a few reasons:

+ I'm encountering cases where it would be handy to: 
    - name different sets of markers that are added interactively. (.e.g by labeling them "bad comparison star" or something).
    - set the marker properties along with the name
+ Better symmetry with `stop_marking`, currently part of the API. Right now one could do either `.is_marking = False` or `.stop_marking()`. 

The first point requires either switching to a method to initiate marking or creating some kind of `Marking` object for specifying marker metadata. 

[...]

PR #60, opened 2019-01-07
Add ability to name markers

mwcraig:
This PR (which really needs an update to the example notebook before it is merged) adds the ability to name sets of markers. If the user does not specify a name, then a default name is used (i.e. the old behavior, essentially).

Markers added interactively, as the result of setting `ImageWidget.is_marking = True` and then clicking, are added to the display with a special name.

There are a couple user-facing changes:

+ `add_markers` now accepts a `marker_name` argument. If, for example, I was was adding all of the APASS stars in the image I might choose the name `'APASS'` for them.
+ `remove_markers` also accepts a `marker_name` argument. To remove my APASS markers I would can `remove_markers(marker_name='APASS')`. Any other markers I had added with a different name (or with no name) would remain.
+ `get_markers` accepts a `marker_name` argument. It also accepts a special value for the marker name, `'all'`, which returns locations for all of the markers.

Ping @eteq because I think this helps address #16. Interactive points are added with the marker name `interactive-markers`. 

If we want the user to choose the name for interactive markers, either because the user doesn't like this name or because they want to do several sets of interactive markers, then we should decide what the API is.


astrowidgets/core.py
```diff
@@ -24,6 +24,9 @@
 # Allowed locations for cursor display
 ALLOWED_CURSOR_LOCATIONS = ['top', 'bottom', None]
 
+# A couple of marker names are for internal use only
+RESERVED_MARKER_SET_NAMES = ['all']
```
pllim: The comment says "a couple" but I only see one defined. I am confused.

astrowidgets/core.py
```diff
@@ -112,6 +115,12 @@ def __init__(self, logger=None, image_width=500, image_height=500,
         # Marker
         self.marker = {'type': 'circle', 'color': 'cyan', 'radius': 20}
         self._marktag = 'marktag'
```
pllim: `self._marktag` is essentially useless now and hence should be removed accordingly from this entire module.

astrowidgets/core.py
```diff
@@ -443,16 +455,28 @@ def get_markers(self, x_colname='x', y_colname='y',
...
+            c_mark = self._viewer.canvas.get_object_by_tag(marker_name)
         except Exception as e:  # No markers
             self.logger.warning(str(e))
             return
 
         image = self._viewer.get_image()
         xy_col = []
 
-        if image.wcs.wcs is None:  # Do not include SkyCoord column
+        if (image is None) or (image.wcs.wcs is None):
```
pllim: Why is `image is None` comparison needed here? One cannot mark on an empty image widget, so at this point of logic, the image has to exist, no?

astrowidgets/core.py
```diff
@@ -500,12 +524,13 @@ def get_markers(self, x_colname='x', y_colname='y',
                 [xy_col[:, 0], xy_col[:, 1], sky_col],
                 names=(x_colname, y_colname, skycoord_colname))
         else:
-            markers_table = Table(xy_col.T, names=(x_colname, y_colname))
```
pllim: Why is `.T` removed? Was it a bug? Has the API changed? I remember adding `.T` but I don't remember why.

astrowidgets/tests/test_image_widget.py
```diff
@@ -0,0 +1,219 @@
...
+    data = np.random.random([300, 300])
+    image.load_array(data)
+    # Send a fake move to the callback. What gets put in the cursor
+    # value should be the event we sent in plus the offset.
+    image.click_center = True
+    data_x = 100
+    data_y = 200
+    image._mouse_move_cb(image._viewer, None, data_x, data_y)
+    output_contents = image._jup_coord.value
+    print(output_contents)
```
pllim: Is this print necessary?

astrowidgets/tests/test_image_widget.py
```diff
@@ -0,0 +1,219 @@
...
+    x_out = re.search(r'X: ([\d\.\d]+)', output_contents)
+    x_out = x_out.groups(1)[0]
+    y_out = re.search(r'Y: ([\d\.\d]+)', output_contents)
+    y_out = y_out.groups(1)[0]
+    assert float(x_out) == data_x + offset
+    assert float(y_out) == data_y + offset
+    # image.print_out.get_state()['outputs']
```
pllim: Remove commented code here.

astrowidgets/core.py
mwcraig: It was a bug...the `.T` raised an error.

astrowidgets/core.py
mwcraig: In practice one *should* always have an image, but it is possible with the current API to add markers in the absence of an image. They don't display, since display only happens when there is an image. I think this came up when I was writing tests...

astrowidgets/tests/test_image_widget.py
mwcraig: nope, deleted

mwcraig:
@pllim -- I think I've addressed the review comments and I've added a notebook demonstrating how marker names work. There are still some rough edges, e.g. some of the attributes I have as private should probably be public, like `ImageWidget._interactive_marker_set_name`

astrowidgets/core.py
pllim: Weird... Pretty sure it worked when I tested it a while ago. Ah, well. :woman_shrugging:

PR #69, opened 2019-02-27
added point shape for marker

michiboo:
Added 2 line for point shape for marker and a example in comment.


michiboo:
Hi! I have converted it to all space.

astrowidgets/core.py
```diff
@@ -417,6 +417,7 @@ def marker(self):
         Marker can be set as follows::
 
             {'type': 'circle', 'color': 'cyan', 'radius': 20}
+            {'type': 'point', 'color': 'green', 'radius':20, style: 'plus/cross' }
```
pllim: I don't think "plus/cross" is valid. You need to pick one.

PR #70, opened 2019-02-28
added point shape for markdown

michiboo:
Hi! the test are added in the function "test_set_markers", and comment are changed. Thank you!

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -417,6 +417,8 @@ class ImageWidget(ipyw.VBox):
         Marker can be set as follows::
 
             {'type': 'circle', 'color': 'cyan', 'radius': 20}
+            {'type': 'cross', 'color': 'green', 'radius': 20}
+            {'type': 'plus', 'color': 'red', 'radius': 20}
 
         """
         return self._marker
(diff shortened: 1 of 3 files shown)


michiboo:
Hi! the test are added in the function  "test_set_markers", and comment are changed. Thank you!

pllim:
Please set up your editor to run `flake8` check on the code and fix PEP 8 warnings. Thanks!

michiboo:
Hi! I have ran flake8 and fix the warning. Thanks for the remainder!

astrowidgets/core.py
```diff
@@ -417,6 +417,7 @@ def marker(self):
         Marker can be set as follows::
 
             {'type': 'circle', 'color': 'cyan', 'radius': 20}
+            {'type': 'point', 'color': 'green', 'radius': 20, 'style': 'plus'}
```
mwcraig: I'd like to make this `'type': 'plus'` and make a separate marker whose type is `'cross'` for the other ginga point option. 

What you have done is perfectly reasonable in that it follows how ginga defines the `Point` shape; since the goal in this package is to eventually provide a few different backends, though, I'd like to keep the point properties as uniform as possible and the names as explicit as possible.

astrowidgets/core.py
```diff
@@ -426,6 +427,8 @@ def marker(self, val):
         marker_type = val.pop('type')
         if marker_type == 'circle':
             self._marker = functools.partial(self.dc.Circle, **val)
+        elif marker_type == 'point':
```
mwcraig: This would become two marker types, `plus` and `cross`

michiboo: Hi I have made the changes as you requested! If there is anything wrong please let me know.

mwcraig:
Looks good, thanks!

PR #71, opened 2019-03-01
finished test_reset_marks

michiboo:
Hi! I finished the test_reset_marks. Also I am wondering is it better to accumulate more changes before open a pull request?

### DIFF
--- a/astrowidgets/tests/test_api.py
+++ b/astrowidgets/tests/test_api.py
@@ -107,8 +107,8 @@ def test_reset_markers():
     image.add_markers(table, x_colname='x', y_colname='y',
                       skycoord_colname='coord', marker_name='test2')
     image.reset_markers()
-    assert image._viewer.canvas.tags.get('test') is None
-    assert image._viewer.canvas.tags.get('test2') is None
+    assert image.get_markers(marker_name='test') is None
+    assert image.get_markers(marker_name='test2') is None
 
 
 def test_remove_markers():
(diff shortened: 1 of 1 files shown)


pllim:
@michiboo , there is no minimum size limit on a PR. I had merged one-liner before. As long as this atomically address the issue, it should be fine. Thanks!

astrowidgets/tests/test_api.py
```diff
@@ -100,8 +100,15 @@ def test_reset_markers():
...
-
-    # TODO add test of actually removing markers...
+    table = Table(data=np.random.randint(0, 100, [5, 2]),
+                  names=['x', 'y'], dtype=('int', 'int'))
+    image.add_markers(table, x_colname='x', y_colname='y',
+                      skycoord_colname='coord', marker_name='test')
+    image.add_markers(table, x_colname='x', y_colname='y',
+                      skycoord_colname='coord', marker_name='test2')
+    image.reset_markers()
+    assert image._viewer.canvas.tags.get('test') is None
```
mwcraig: Could you please change these to check the results of calls to `image.get_markers` instead of calling the underlying ginga viewer? I'd like to try to make sure that the tests rely as much as possible on the public API rather than the internals of the particular viewer.

michiboo: Hi! sure! I have updated the code, thanks!

mwcraig:
Thanks, will try to look at the rest tomorrow!

PR #74, opened 2019-03-07
Update installation instruction and notes on "index.rst"

kakirastern:
To update installation instruction with a modified command, plus to add a note to accompany it to warn users to check and use the appropriate pip version to install astrowidgets...

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -16,12 +16,15 @@ Installation
 
 Currently, ``astrowidgets`` are only available via GitHub::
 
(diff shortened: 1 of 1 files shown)


kakirastern:
Yeah, I reckon some other people probably had tried it and knew about the error but they had other ways out, so didn't report it.

mwcraig:
Looks good, merging! That install line used to work, but I guess pip evolves pretty rapidly these days....

PR #75, opened 2019-03-09
add parameter type checking for constructor

michiboo:
Hi I have added type checking for Image widget constructor.

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -32,8 +32,6 @@ class ImageWidget(ipyw.VBox):
     """
     Image widget for Jupyter notebook using Ginga viewer.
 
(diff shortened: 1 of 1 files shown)


michiboo:
I will try do some Google search on this!

michiboo:
So type hinting is only for static code analysis and do not raise exception.

mwcraig:
I'm not sure that the intent here was to check that the user passed in a valid type....my best guess for what we meant by `.. todo:: Any property passed to constructor has to be valid keyword.` is that we were thinking the signature of `__init__` would be `__init__(*arg, *kwargs)` and that we would need to let the user know if one of the keyword args they passed in was unrecognized.

PR #78, opened 2019-03-21
Add a generic load_file method

### LINKED ISSUE #64: ENH: Add interface to Ginga's ASDF loader
I made an attempt at spacetelescope/dat_pyinthesky#7 during a "hack day" specifically for JWST but it can be generalized for just the generic ASDF format (https://asdf.readthedocs.io/en/latest/). Relevant snippet from the hack day:

```python
    # TODO: Need to generalize this method before moving it
    #       to astrowidgets package.
    def load_jwst_asdf(self, filename):
        """Load ASDF extension from JWST Level 2 data.

        .. note::

            This needs to use ``'astropy_ape14'`` WCS module in Ginga.
            This also currently needs ``jwst`` pipeline to be installed.

        Parameters
        ----------
        filename : str
            Filename of JWST Level 2 file.
            It is a FITS file with ASDF-in-FITS extension.

        """
        import asdf

        image = AstroImage(logger=self.logger)

        with asdf.open(filename) as asdf_f:
            asdf_f['wcs'] = asdf_f['meta']['wcs']
            image.load_asdf(asdf_f, data_key='data')

        self._viewer.set_image(image)
```

Ginga's ASDF loader was implemented in ejeschke/ginga#719 and will be released with Ginga 3.0.

pllim:
This would fix #64 indirectly, as theoretically I could implement whatever Ginga file handler I want outside of the widget, register it with Ginga, and then simply use this to open my custom file format.

https://github.com/ejeschke/ginga/blob/bbfe6455d302266c1bfd3f01bd386fbc2494f3db/ginga/AstroImage.py#L186

I can add a test if you think this is an acceptable way forward. I am open to discussions. Thanks!

Depends on: ejeschke/ginga#764

cc @ejeschke

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -251,8 +251,11 @@ class ImageWidget(ipyw.VBox):
             Extra keywords to be passed into the viewer's file handler.
 
         """
-        image = AstroImage(logger=self.logger)
-        image.load_file(filename, **kwargs)
+        from ginga.util import loader
+
+        image = loader.load_data(filename, logger=self.logger, **kwargs)
+        if not isinstance(image, AstroImage):
+            raise ValueError("Default index in '{}' does not reference an image".format(filename))
         self._viewer.set_image(image)
 
     def load_fits(self, fitsorfn, numhdu=None, memmap=None):
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -237,6 +237,24 @@ class ImageWidget(ipyw.VBox):
 #         from IPython.display import display
 #         return display(self._widget)
 
+    def load_file(self, filename, **kwargs):
+        """
+        Load a given file into the viewer. This relies on the viewer
+        to identify the best file handler to use.
+
+        Parameters
+        ----------
+        filename : str
+            File name.
+
+        kwargs : dict
+            Extra keywords to be passed into the viewer's file handler.
+
+        """
+        image = AstroImage(logger=self.logger)
+        image.load_file(filename, **kwargs)
+        self._viewer.set_image(image)
+
     def load_fits(self, fitsorfn, numhdu=None, memmap=None):
         """
         Load a FITS file into the viewer.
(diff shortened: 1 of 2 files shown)


pllim:
Thanks for the review, @mwcraig ! I added a test, as requested.

pllim:
p.s. Do I also have to open an issue in @eteq 's API thingy?

astrowidgets/core.py
```diff
@@ -237,6 +237,24 @@ def _mouse_click_cb(self, viewer, event, data_x, data_y):
...
+
+        Parameters
+        ----------
+        filename : str
+            File name.
+
+        kwargs : dict
+            Extra keywords to be passed into the viewer's file handler.
+
+        """
```
ejeschke: @pllim, you might want to use:
```python
from ginga.util import loader
image = loader.load_data(filename, **kwargs)
if not isinstance(image, AstroImage):
    raise ValueError("Default index in '{}' does not reference an image".format(filename))
self._viewer.set_image(image)
```

ejeschke: Do you have a parameter for specifying HDU or some kind of index in the astrowidget load API?  Because you could pass it in to `idx` in this `load_data` call.

pllim: This new method is meant more for ASDF than for FITS, though ideally it would work on either one if the Ginga file loader registry is set up properly.

pllim:
@ejeschke , test is failing because `ginga.util.loader` is not in the stable version yet. I guess we'll have to wait for a Ginga release?

mwcraig:
> p.s. Do I also have to open an issue in @eteq 's API thingy?

May as well, yes

pllim:
We agreed to do refactor this file loading API and have it be backend agnostic. See [notes from Astropy Coordination Meeting 2023](https://docs.google.com/document/d/1sBImbw1p4mM3VYWGQJDRbj2N3uqP5Ed_Wfb8BDjJj2o/view).

PR #82, opened 2019-06-21
Implement the remaining API for the ginga viewer

### LINKED ISSUE #16: select_points / session-ful marking

mwcraig:
This started as a refactor of the marking interface, and does fix #16, but ended up also implementing the remaining parts of the API that hadn't been implemented yet. 

A little bit of documentation and we would be ready for an alpha release, I think?

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -112,6 +112,12 @@ class ImageWidget(ipyw.VBox):
         self.canvas.enable_draw(True)
         self.canvas.enable_edit(True)
 
+        bind_map = self._viewer.get_bindmap()
+        # Set up right-click and drag adjusts the contrast
+        bind_map.map_event(None, (), 'ms_right', 'contrast')
+        # Shift-right-click restores the default contrast
+        bind_map.map_event(None, ('shift',), 'ms_right', 'contrast_restore')
+
         # Marker
         self.marker = {'type': 'circle', 'color': 'cyan', 'radius': 20}
         # Maintain marker tags as a set because we do not want
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -72,8 +72,6 @@ class ImageWidget(ipyw.VBox):
                 warnings.warn('install opencv or set use_opencv=False')
 
         self._viewer = EnhancedCanvasView(logger=logger)
-        self._is_marking = False
-        self._click_center = False
 
         self._pixel_offset = pixel_coords_offset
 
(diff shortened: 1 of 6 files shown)


astrowidgets/core.py
```diff
@@ -384,28 +403,50 @@ def is_marking(self):
...
-    @is_marking.setter
-    def is_marking(self, val):
-        if not isinstance(val, bool):
-            raise ValueError('Must be True or False')
-        elif self.click_center and val:
-            raise ValueError('Cannot set to True while in click-center mode')
-        self._is_marking = val
-
-    def stop_marking(self, clear_markers=True):
+    # @is_marking.setter
```
pllim: I'd rather this is deleted then just commented. We can always check `git history` if we need to see it again, so there is no need to carry this around?

mwcraig: oops, forgot about that, fixing

PR #86, opened 2019-06-27
ENH: Update CI matrix

### LINKED ISSUE #83: Add ginga dev to test matrix

pllim:
* Fix #83 by adding Ginga dev to Astropy dev job.
* Follow the lead of astropy/astropy#8891 and remove Numpy pre-release job. Any breakage caused by unreleased Numpy will be likely caught upstream first anyway. 🤞 
* Remove warnings from setup process.
* Update Appveyor to pip install Ginga (could not find in `conda`) and use Python 3.7.
* Improve pytest header to report dependency versions that we care about.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ ignore = E501
 exclude = setup_package.py,conftest.py,__init__.py
 
 [metadata]
-package_name = astrowidgets
+name = astrowidgets
 description = Widgets for the Jupyter notebook and JupyterLab
 long_description =
 author = Astropy Developers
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -82,10 +81,6 @@
 # modify distutils' behavior.
 cmdclassd = register_commands(PACKAGENAME, VERSION, RELEASE)
 
-# Freeze build information in version.py
-generate_version_py(PACKAGENAME, VERSION, RELEASE,
```
pllim: @astrofrog or @bsipocz , is it safe to just remove this? Or do I have to replace this with something else?

bsipocz: I would use whatever is in the template, both for this and `setup.cfg`
`version = generate_version_py()`

setup.cfg
```diff
@@ -26,7 +26,7 @@ ignore = E501
 exclude = setup_package.py,conftest.py,__init__.py
 
 [metadata]
-package_name = astrowidgets
+name = astrowidgets
```
mwcraig: Looks like this is causing https://travis-ci.org/astropy/astrowidgets/jobs/551127071#L1442

pllim: Ops. Fixed!

pllim:
Travis job is running but it is not reporting back here. I am also unable to log into `travis.org`. :scream:

pllim:
Travis CI log at https://travis-ci.org/astropy/astrowidgets/builds/551360172 .

There are some complains on Appveyor but it is a battle for another day. See astropy/ci-helpers#398 .

PR #91, opened 2019-08-02
Add some docs

mwcraig:
This is nowhere near complete but better than it was....

### DIFF
--- a/docs/astrowidgets/index.rst
+++ b/docs/astrowidgets/index.rst
@@ -1,18 +1,67 @@
-Image Widgets
-=============
+Image widget for Jupyter Lab/notebook
+=====================================
+
+Getting started
+---------------
+
+Make a viewer
++++++++++++++
+
+The snippet below is all you need to make an image widget. The widget is part
+of the `ipywidgets framework <https:ipywidgets.rtfd.io>`_ so that it can
+be easily integrated with other controls::
+
+    >>> from astrowidgets import ImageWidget
+    >>> image = ImageWidget()
+    >>> display(image)
+
+Loading an image
+++++++++++++++++
+
+An empty viewer is not very useful, though, so load some data from a FITS
+file::
+
+    >>> image.load_fits('https://zenodo.org/record/3356833/files/kelt-16-b-S001-R001-C084-r.fit.bz2?download=1')
+
+The image widget can also load a numpy array via
+`~astrowidgets.ImageWidget.load_array`. It also understands astropy
+`~astropy.nddata.NDData` objects; load them via
+`~astrowidgets.ImageWidget.load_data`.
+
+Navigation
+++++++++++
+
+In the default configuration basic navigation is done using these controls:
+
+* scroll to pan
+* use ``+``/``-`` to zoom in/out (cursor must be over the image for this to work)
+* right-click and drag to change contrast ds9-style
+
+API
++++
+
+One important design goal is to make all functionality available by a compact,
+clear API. The `target API <https://github.com/eteq/nb-astroimage-api>`_ still
+needs a few features (e.g. blink), but much of it is already implemented.
+
+The API-first approach means that manipulating the view programmatically is straightforward. For example, centering on the position of the object Kelt-16 and zooming in to 8x the natural pixel scale is straightforward::
+
+    >>> from astropy.coordinates import SkyCoord
+    >>> image.center_on(SkyCoord.from_name('kelt-16'))
+    >>> image.zoom_level = 8
+
+A more detailed description of the interface and the :ref:`api-docs` are available.
+
+.. toctree::
+    :maxdepth: 2
+
+    api.rst
 
 Example Notebooks
 -----------------
 
-* `astrowidgets using Ginga <https://github.com/astropy/astrowidgets/blob/master/example_notebooks/ginga_widget.ipynb>`_
+* `astrowidgets using the Ginga backend <https://github.com/astropy/astrowidgets/blob/master/example_notebooks/ginga_widget.ipynb>`_
+* `Using named markers to keep track of logically related markers <https://github.com/astropy/astrowidgets/blob/master/example_notebooks/named_markers.ipynb>`_
+* `Demonstration of GUI interactions <https://github.com/astropy/astrowidgets/blob/master/example_notebooks/gui_interactions.ipynb>`_
 
 
-Under Construction
-------------------
-
-User-facing documentation goes here:
-
-* How do you use an existing widget?
-* How do you expand on an existing widget?
-* How do you develop a new widget from metaclass?
-* Space cats?
--- a/docs/astrowidgets/index.rst
+++ b/docs/astrowidgets/index.rst
@@ -19,7 +19,8 @@ Loading an image
 ++++++++++++++++
 
 An empty viewer is not very useful, though, so load some data from a FITS
-file::
+file. The FITS file at the link below is an image of the field of the exoplanet
+Kelt-16, and also contains part of the Veil Nebula::
 
     >>> image.load_fits('https://zenodo.org/record/3356833/files/kelt-16-b-S001-R001-C084-r.fit.bz2?download=1')
 
--- a/docs/astrowidgets/index.rst
+++ b/docs/astrowidgets/index.rst
@@ -8,7 +8,7 @@ Make a viewer
 +++++++++++++
 
 The snippet below is all you need to make an image widget. The widget is part
-of the `ipywidgets framework <https:ipywidgets.rtfd.io>`_ so that it can
+of the `ipywidgets framework <https://ipywidgets.rtfd.io>`_ so that it can
 be easily integrated with other controls::
 
     >>> from astrowidgets import ImageWidget
@@ -23,7 +23,7 @@ file::
 
     >>> image.load_fits('https://zenodo.org/record/3356833/files/kelt-16-b-S001-R001-C084-r.fit.bz2?download=1')
 
-The image widget can also load a numpy array via
+The image widget can also load a Numpy array via
 `~astrowidgets.ImageWidget.load_array`. It also understands astropy
 `~astropy.nddata.NDData` objects; load them via
 `~astrowidgets.ImageWidget.load_data`.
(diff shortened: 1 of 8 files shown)


mwcraig:
Done, let me know if you want me to make sure it doesn't end up in the git history.

JuanCab:
> Done, let me know if you want me to make sure it doesn't end up in the git history.

I'm not that concerned, just figured my name didn't need to be part of an error message.

pllim:
Weird... `aggdraw` is throwing an error:
```
Searching for freetype

Reading https://pypi.org/simple/freetype/

No local packages or working download links found for freetype

error: Could not find suitable distribution for Requirement.parse('freetype')
```

pllim:
I think you need to set up the CI to `conda install aggdraw` so that it pulls in `freetype`.

mwcraig:
This is why one is supposed to keep PRs to one topic, isn't it 🤣 

Fix incoming....

docs/astrowidgets/index.rst
```diff
@@ -1,18 +1,67 @@
...
+* right-click and drag to change contrast ds9-style
+
+API
++++
+
+One important design goal is to make all functionality available by a compact,
+clear API. The `target API <https://github.com/eteq/nb-astroimage-api>`_ still
+needs a few features (e.g. blink), but much of it is already implemented.
+
+The API-first approach means that manipulating the view programmatically is straightforward. For example, centering on the position of the object Kelt-16 and zooming in to 8x the natural pixel scale is straightforward::
```
pllim:
```suggestion
The API-first approach means that manipulating the view programmatically is straightforward.
For example, centering on the position of the object, Kelt-16, and zooming in to 8x the natural
pixel scale is straightforward::
```

docs/index.rst
```diff
@@ -5,99 +5,22 @@ astrowidgets
...
-nodejs
-^^^^^^
-
-In Windows 7, ``conda install -c conda-forge nodejs`` might throw an
-``IOError``. The workaround for this is to install ``yarn`` and ``nodejs``
-outside of ``conda`` from https://yarnpkg.com (the stable version) and
-https://nodejs.org (the LTS version), respectively.
+    ``astrowidgets`` is currently in very early development.
+    There may be significant changes in the API, though we will do
+    our best to follow semantic versioning
```
pllim:
```suggestion
    our best to follow semantic versioning.
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+    conda install -c conda-forge astrowidgets nodejs
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager
+    # Optional, but avoids a rebuild when you open lab
+    jupyter labextension build
+
+.. note::
+
+    If you are using a virtual environment and
+    are experiencing problems with the installation, be sure to check and use
+    the suitable pip version. Also, if you are having problem even with the following
```
pllim:
```suggestion
    the suitable ``pip`` version. Also, if you are having problem even with the following
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+    dependencies installed, please make sure you have the most up-to-date versions.
+    Upgrade dependencies when and where necessary.
+
+The following dependencies are needed *in the kernel of execution*
+to use ``astrowidgets`` in either Jupyter Lab or Notebook. They should be installed automatically when you install astrowidgets:
```
pllim:
```suggestion
to use ``astrowidgets`` in either Jupyter Lab or Notebook. They should be installed
automatically when you install astrowidgets:
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+
+* ``python >= 3.6``
```
pllim:
```suggestion
* ``python>=3.6``
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+* ``numpy``
+* ``astropy``
+* ``ipywidgets>=7.5``
+* ``ipyevents >= 0.6.3``
```
pllim:
```suggestion
* ``ipyevents>=0.6.3``
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+* ``ginga>=2.7.1``
+* ``pillow``
+* ``freetype``
+* ``aggdraw``
+* ``jupyterlab>=1``
+* ``opencv`` (optional, not installed by default)
+* ``nodejs`` (only if you plan to use Jupyter Lab, not installed by default)
+
+After installing dependencies, run::
```
pllim:
```suggestion
After installing dependencies, for Jupyter Lab, run::
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager
+
+For those using ``conda``, dependencies from the ``conda-forge`` channel
+should be sufficient unless stated otherwise.
+
+Using OpenCV
+------------
+
+If you wish to use `OpenCV <https://docs.opencv.org/master/index.html>`_ to handle the
+drawing in ginga you have two options:
```
pllim:
```suggestion
drawing in Ginga, you have two options:
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+
+Install OpenCV with pip
+^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are using pip it looks like the best option is to use the
+`opencv-python` package, which provides pre-built binaries of most of OpenCV::
```
pllim:
```suggestion
``opencv-python`` package, which provides pre-built binaries of most of OpenCV::
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+"unofficial" so you should probably read about the project before using.
+
+Install OpenCV with conda
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This should work on conda::
+
+    conda install -c conda-forge opencv
+
+If, after installing opencv, you get a warning like this::
```
pllim:
```suggestion
If, after installing ``opencv``, you get a warning like this::
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+
+    astrowidgets/core.py:72: UserWarning: install opencv or set use_opencv=False
+    warnings.warn('install opencv or set use_opencv=False')
+
+then you should try installing a newer version of freetype::
```
pllim:
```suggestion
then you should try installing a newer version of ``freetype``::
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+
+    conda install 'freetype\>=2.10'
+
+For more details, see `this discussion of opencv and astrowidgets
+<https://github.com/astropy/astrowidgets/issues/90>`_.
+
+Widget with Ginga Toolkit
```
pllim:
```suggestion
Widget with Ginga toolkit
```

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+-------------------------
+
+.. note::
+
+    For vectorized drawing in ``aggdraw``, you can clone
```
pllim: Is this still necessary, @ejeshcke?

Also, one section with nothing but one note seems incomplete somehow. Any plans to add more text in this section, @mwcraig ?

docs/install.rst
```diff
@@ -0,0 +1,127 @@
...
+    https://github.com/ejeschke/aggdraw/ and install its ``vectorized-drawing``
+    branch from source.
+
+
+Notes for Windows Users
```
pllim:
```suggestion
Notes for Windows users
```

setup.cfg
```diff
@@ -37,7 +39,7 @@ edit_on_github = True
 github_project = astropy/astrowidgets
 # install_requires should be formatted as a comma-separated list, e.g.:
 # install_requires = astropy, scipy, matplotlib
-install_requires = astropy, ginga, pillow, ipywidgets, ipyevents>=0.4.0
+install_requires = astropy, ginga, pillow, ipywidgets>=7.5, ipyevents>=0.6.3, jupyterlab>=1, aggdraw
```
pllim: I feel like requiring jupyterlab on install is overkill, given that the widget can run without it?

docs/install.rst
mwcraig: @ejeschke -- is this workaround still necessary?

setup.cfg
mwcraig: On the other hand, end-of-life of classic notebook isn't too far off (hard date has not been announced as far as I know), and it is an easy install (wheels available). Can remove it if you prefer though, but will need to update the install instructions then.

pllim: Hmm... I guess it doesn't hurt to leave it in for now.

docs/install.rst
```diff
@@ -0,0 +1,128 @@
...
+* ``astropy``
+* ``ipywidgets>=7.5``
+* ``ipyevents>=0.6.3``
+* ``ginga>=2.7.1``
+* ``pillow``
+* ``freetype``
+* ``aggdraw``
+* ``jupyterlab>=1``
+* ``opencv`` (optional, not installed by default)
+* ``nodejs`` (only if you plan to use Jupyter Lab, not installed by default)
```
pllim: So, I still thinks the instruction is a bit confusing because `jupyterlab` is installed regardless as a requirement but here the text suggests that it is an optional dependency.

docs/install.rst
ejeschke: Let me take a look...

ejeschke: Yes, and it currently works only for circles.

ejeschke: @mwcraig, are you providing a vectorized `Circle` class?  I don't think we added one to Ginga yet.  Didn't we do this as an experimental branch?  Not sure we should mention it until it is available.

ejeschke: I do have a branch of `aggdraw` that supports vectorized drawing of objects that Ginga supports.  My hope was that we could refactor the agg renderer in Ginga so that it would be unecessary to create vector objects; i.e. you could just add as many different objects to the Ginga canvas and it would draw them using the vectorized renderer.

mwcraig: > @mwcraig, are you providing a vectorized Circle class? I don't think we added one to Ginga yet. Didn't we do this as an experimental branch? Not sure we should mention it until it is available.

Sounds like this should be removed from the documentation for now. I've found performance to be pretty good as long as markers are added in pixel coordinates. I switched the astrowidgets code to convert world coordinates to pixel before marking here https://github.com/astropy/astrowidgets/commit/7c9fb3694029144c5c5be5079ee6ee9e53547d0d

ejeschke: Yeah, you will save a round of WCS conversions on redraws if you do that.

docs/install.rst
mwcraig: Fixed, I think...

docs/astrowidgets/index.rst
```diff
@@ -1,18 +1,67 @@
...
+    >>> image = ImageWidget()
+    >>> display(image)
+
+Loading an image
+++++++++++++++++
+
+An empty viewer is not very useful, though, so load some data from a FITS
+file::
+
+    >>> image.load_fits('https://zenodo.org/record/3356833/files/kelt-16-b-S001-R001-C084-r.fit.bz2?download=1')
```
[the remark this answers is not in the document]
mwcraig: done

mwcraig:
@pllim -- I think I've addressed all of your comments.

mwcraig:
That RTD thing is amazing...

PR #92, opened 2019-08-03
Drop appveyor in favor of travis and increase minimum python version to 3.6

mwcraig:
This drops appveyor for obvious reasons and raises the minimum python version in prep for the first release since python 3.8 is almost out.

And I really like f-strings....


mwcraig:
Merging; the appveyor failure is because there is no setup file 😀

PR #96, opened 2019-09-12
[EXP] Proof of concept using Ginga with regions support

pllim:
This adds `regions` support for `astrowidgets` with Ginga backend. Companion of ejeschke/ginga#787.

The API is not in @eteq's original API design, so discussions welcome. I am also not sure how to tie this into the existing markers machinery without hackery; maybe @ejeschke has suggestions.


astrowidgets/core.py
```diff
@@ -619,6 +619,29 @@ def _validate_marker_name(self, marker_name):
...
+            ``regions`` shapes.
+
+        """
+        from ginga.util import ap_region
+
+        # Regions will be appended if other regions were previously
+        # added in another call.
+        # TODO: How to get or set the canvas name?
+        for reg in regions:
+            ap_region.add_region(self._viewer.canvas, reg)
```
pllim: Is this too lazy? Should I be using a lower-level Ginga function that is `astropy_region_to_ginga_canvas_object`?

ejeschke:
I think a main question to ask is what objects does this API support plotting. `regions` objects go a bit beyond the basic idea of "markers".  And then there are other interesting things that you could plot if you want to expose them to the user, like compasses, rulers, WCS grid or color bars.

pllim:
Thanks for the input, @ejeschke . I will have to defer to @mwcraig and @eteq on their vision for marker/region support. For `astrowidgets` to be usable across different GUI packages, it cannot lock too much into Ginga specific functionalities, but I don't know where to draw the line as I am a Ginga co-maintainer myself.

mwcraig:
> but I don't know where to draw the line as I am a Ginga co-maintainer myself.

💯 on that pun...

astrowidgets/core.py
```diff
@@ -619,6 +619,32 @@ def _validate_marker_name(self, marker_name):
...
+            Name to assign the markers created from regions.
+
+        """
+        from ginga.util import ap_region
+
+        # Regions will be appended if other regions were previously
+        # added in another call.
+        # TODO: How to get or set the canvas name?
+        for reg in regions:
+            ap_region.add_region(self._viewer.canvas, reg, tag=marker_name)
```
mwcraig: This fails for all regions after the first the error below. We get around this in `add_markers` by creating a compound object. We could (maybe?) do that here, or generate a name for each region or something like that.


```python traceback
---------------------------------------------------------------------------
CanvasError                               Traceback (most recent call last)
<ipython-input-10-a84f619487a4> in <module>
----> 1 w.add_regions(ds9_pixel_regions)

~/development/astronomy/astrowidgets/astrowidgets/core.py in add_regions(self, regions, marker_name)
    644         # TODO: How to get or set the canvas name?
    645         for reg in regions:
--> 646             ap_region.add_region(self._viewer.canvas, reg, tag=marker_name)
    647 
    648     def add_markers(self, table, x_colname='x', y_colname='y',

~/miniconda3/envs/awid-ging-dev/lib/python3.7/site-packages/ginga/util/ap_region.py in add_region(canvas, r, tag, redraw)
    164         tag = obj.get_data('name')
    165     if obj is not None:
--> 166         canvas.add(obj, tag=tag, redraw=redraw)
    167         return obj
    168 

~/miniconda3/envs/awid-ging-dev/lib/python3.7/site-packages/ginga/canvas/CanvasMixin.py in add(self, obj, tag, tagpfx, belowThis, redraw)
     51             # user supplied a tag
     52             if tag in self.tags:
---> 53                 raise CanvasError("Tag already used: '%s'" % (tag))
     54         else:
     55             if tagpfx:

CanvasError: Tag already used: '_astropy_regions'
```

mwcraig: This is where we handle tag names in markers: https://github.com/astropy/astrowidgets/pull/96/files#diff-47161bfaadb2c20c9c4f939f18fb08aaL697-L710

mwcraig:
I played around a bit more and the thing we do for markers -- create a compound object and add that to the canvas -- seems to work here. I'll push a couple commits to your branch that implement this.

pllim:
This will need a rebase. @mwcraig , if you think this is acceptable, feel free to merge when conflict is resolved, unless you are waiting on @eteq ?

mwcraig:
Close/reopen to re-run tests

pllim:
I forgot we had this. I am so glad it still works! 😸

PR #98, opened 2019-09-27
Fix some edge cases in using named marker sets

### LINKED ISSUE #97: get_markers does not handle empty tables well
User @isobelsnellenberger reported a couple of issues with the `get_markers` method. 

In the code snippets below, assume an `ImageWidget` called `iw` has been created.

1. When you try to get markers using a name for which there are no markers, e.g. `iw.get_markers(marker_name='no markers by this name')` the only output is the name with no markers, e.g. ` 'no markers by this name'`, which is not helpful. This can happen two ways: a) the marker name is exists in the internal list of names but no markers are on it (e.g. `iw.start_marking(marker_name='new markers')` but no markers have been added interactively) or  the `marker_name` doesn't exist in the internal list at all.
2. If you get all markers with  `iw.get_markers(marker_name='all')` and there is is a marker name in the internal list that has no markers in it, then the an exception is raised when the code tries to iterate over the rows in a table that does not yet exist (see traceback below).

PR incoming soon to fix this.

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-2b30fa1bc578> in <module>
----> 1 all_table = iw.get_markers(marker_name='all')

~/conda-main/envs/ast366xx/lib/python3.7/site-packages/astrowidgets/core.py in get_markers(self, x_colname, y_colname, skycoord_colname, marker_name)
    532                                          marker_name=name)
    533                 try:
--> 534                     coordinates.extend(c for c in table[skycoord_colname])
    535                 except KeyError:
    536                     pass

TypeError: 'NoneType' object is not subscriptable
```

mwcraig:
This fixes #97 by handling these cases:

+ User asks for a marker name that the image widget does not know about (raises a `ValueError`)
+ User asks for a marker name which the widget knows about but has no markers in it (raise a `RuntimeWarning` with a clear message)
+ User asks for all markers and one of the marker sets has no markers in it (don't crash, just ignore the empty marker set).

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -530,6 +530,10 @@ class ImageWidget(ipyw.VBox):
                                          y_colname=y_colname,
                                          skycoord_colname=skycoord_colname,
                                          marker_name=name)
+                if table is None:
+                    # No markers by this name, skip it
+                    continue
+
                 try:
                     coordinates.extend(c for c in table[skycoord_colname])
                 except KeyError:
@@ -553,7 +557,8 @@ class ImageWidget(ipyw.VBox):
 
         try:
             c_mark = self._viewer.canvas.get_object_by_tag(marker_name)
-        except Exception:  # No markers
+        except Exception:
+            # No markers in this table. Issue a warning and continue
             warnings.warn(f"Marker set named '{marker_name}' is empty",
                           category=RuntimeWarning)
             return None
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -553,9 +553,10 @@ class ImageWidget(ipyw.VBox):
 
         try:
             c_mark = self._viewer.canvas.get_object_by_tag(marker_name)
-        except Exception as e:  # No markers
-            self.logger.warning(str(e))
-            return
+        except Exception:  # No markers
+            warnings.warn(f"Marker set named '{marker_name}' is empty",
+                          category=RuntimeWarning)
+            return None
 
         image = self._viewer.get_image()
         xy_col = []
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -545,6 +545,12 @@ class ImageWidget(ipyw.VBox):
 
             return stacked
 
+        # We should always allow the default name. The case
+        # where that table is empty will be handled in a moment.
+        if (marker_name not in self._marktags and
+                marker_name != self._default_mark_tag_name):
+            raise ValueError(f"No markers named '{marker_name}' found.")
+
         try:
             c_mark = self._viewer.canvas.get_object_by_tag(marker_name)
         except Exception as e:  # No markers
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -560,7 +560,7 @@ class ImageWidget(ipyw.VBox):
         except Exception:
             # No markers in this table. Issue a warning and continue
             warnings.warn(f"Marker set named '{marker_name}' is empty",
-                          category=RuntimeWarning)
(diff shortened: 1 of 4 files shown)


astrowidgets/core.py
```diff
@@ -545,11 +549,19 @@ def get_markers(self, x_colname='x', y_colname='y',
...
+
         try:
             c_mark = self._viewer.canvas.get_object_by_tag(marker_name)
-        except Exception as e:  # No markers
-            self.logger.warning(str(e))
-            return
+        except Exception:
+            # No markers in this table. Issue a warning and continue
+            warnings.warn(f"Marker set named '{marker_name}' is empty",
+                          category=RuntimeWarning)
```
pllim: Is `RuntimeWarning` the best one to use? According to https://docs.python.org/3/library/warnings.html#warning-categories , it is for "warnings about dubious runtime features."

pllim:
Failure is real:
```
________________________ test_unknown_marker_name_error ________________________
    def test_unknown_marker_name_error():
        """
        Regression test for https://github.com/astropy/astrowidgets/issues/97
    
        This particular test checks that getting a marker name that
        does not exist raises an error.
        """
        iw = ImageWidget()
        bad_name = 'not a real marker name'
        with pytest.raises(ValueError) as e:
            iw.get_markers(marker_name=bad_name)
    
>       assert f"No markers named '{bad_name}'" in str(e)
E       assert "No markers named 'not a real marker name'" in '<ExceptionInfo ValueError tblen=2>'
E        +  where '<ExceptionInfo ValueError tblen=2>' = str(<ExceptionInfo ValueError tblen=2>)
astrowidgets/tests/test_image_widget.py:259: AssertionError
```

astrowidgets/tests/test_image_widget.py
```diff
@@ -215,3 +242,61 @@ def test_get_marker_with_names():
...
+
+    This particular test checks that getting a marker name that
+    does not exist raises an error.
+    """
+    iw = ImageWidget()
+    bad_name = 'not a real marker name'
+    with pytest.raises(ValueError) as e:
+        iw.get_markers(marker_name=bad_name)
+
+    assert f"No markers named '{bad_name}'" in str(e)
```
pllim:
```suggestion
    assert f"No markers named '{bad_name}'" in str(e.value)
```

mwcraig: thanks, have upgraded my local pytest...

astrowidgets/core.py
mwcraig: I wasn't sure which one to go with....I picked runtime because that is what numpy does for things like underflow/overflow errors. Could change it to `UserWarning`.

mwcraig:
flake8 is being difficult...I'm going to disable W503 (PEP8 leans towards the line break before the binary operator)

fish___Users_mcraig_Development_astrowidgets__fish__and_core_py_—_astrowidgets

pllim:
Yeah, the line break thingy is where I draw my line and start shushing it. 😉

astrowidgets/core.py
pllim: Empty marker list isn't quite the same as under/overflow. I would prefer `UserWarning` but I'll leave the final decision to you.

setup.cfg
```diff
@@ -24,7 +24,7 @@ offline = True
 
 [flake8]
 # E501: line too long
-ignore = E501
+ignore = E501,W503
```
pllim: For future clarity, please also add a comment after L26 to explain what W503 is.

PR #99, opened 2019-10-14
Add test and fix for case of adding single SkyCoord marker

mwcraig:
Thanks to @isobelsnellenberger for finding this bug...

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -696,7 +696,7 @@ class ImageWidget(ipyw.VBox):
                                                            0)
             # In the event a *single* marker has been added, coord_x and coord_y
             # will be scalars. Make them arrays always.
-            if np.isscalar(coord_x):
+            if np.ndim(coord_x) == 0:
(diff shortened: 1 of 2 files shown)


astrowidgets/core.py
```diff
@@ -694,6 +694,13 @@ def add_markers(self, table, x_colname='x', y_colname='y',
             coord_x, coord_y = image.wcs.wcs.all_world2pix(coord_val.ra.deg,
                                                            coord_val.dec.deg,
                                                            0)
+            # In the event a *single* marker has been added, coord_x and coord_y
+            # will be scalars. Make them arrays always.
+            try:
+                len(coord_x)
+            except TypeError:
```
pllim: `except` is computationally expensive if it is invoked a lot. Why not use `np.isscalar` as a check?

mwcraig: Mostly because I forgot that `np.isscalar` existed...changing.

PR #105, opened 2020-01-08
DOC: Clarify ipywidgets install

pllim:
This is in https://ipywidgets.readthedocs.io/en/latest/user_install.html but it is not obvious to me what was happening until @stargaser (thanks, David!) pointed it out, so I thought might as well state this in the install doc?


pllim:
Still does not work for me... Never mind.

PR #106, opened 2020-01-09
[EXP] WCS locking in multiple widgets

### LINKED ISSUE #62: Feature Request: WCS locking between different image displays in the same notebook
~~Currently, this works for "pixel locking"~~, but what about "WCS locking"; i.e., ImageWidget A and ImageWidget B (both displayed in the same notebook) have similar pointing but different central RA and DEC and/or rotation, what does it mean to lock their WCS when one pan/zoom on one of them? How is it accomplished? Do we need to rotate one of them so they both have the same orientation?
```python
import ipywidgets

mylink = ipywidgets.widgets.jslink((w._jup_img, 'value'), (w2._jup_img, 'value'))
```

UPDATE: `jslink` value does not really do what I want...

xref spacetelescope/dat_pyinthesky#7

pllim:
Fix #62 

* `linked_wcs_2_windows.ipynb` addresses the **2 linked windows with "a hundred" candidates to load for each** use case in #62 
  * Caveat: I don't have a hundred files on disk to test this with, but the machinery is in place if you want to try load 100 files into the notebook...
  * Caching was added to this PR in subsequent commit.
* `linked_wcs_blink.ipynb` addresses the **1 window to "blink" 2 candidates manually with a "hot key"** use case in #62 (caching using `Datasrc` is also added, xref #108)
* `linked_wcs_12_windows.ipynb` addressed the **12 linked windows with one candidate loaded for each** use case in #62 (also with caching)
  * Caveat: I only have 3 files, so I repeatedly use them to fill all 12 viewers.

cc @hcferguson and @larrybradley

Data files used available from https://stsci.app.box.com/folder/3381606676 (special permission required).

**TODO**

- [x] Load simulated calibrated dithered NIRCam images provided by @bhilbert4 (ASDF in FITS).
- [x] Implement caching for `linked_wcs_2_windows.ipynb`.
- [x] See if there is anything we can do to mitigation a visual jump when blinking in `linked_wcs_blink.ipynb`. (Nope. But maybe not critical as it gives visual queue that new image is being displayed.)
- [x] Stretch goal: Implement a third notebook addressing "12 linked windows with one candidate loaded for each" use case in #62 

### linked_wcs_blink.ipynb

Screenshot_blink_wcs

### linked_wcs_2_windows.ipynb

Screenshot_link_2_wcs

### linked_wcs_12_windows.ipynb

Screenshot_12_monkeys

### DIFF
--- a/example_notebooks/linked_wcs_12_windows.ipynb
+++ b/example_notebooks/linked_wcs_12_windows.ipynb
@@ -1,0 +1,247 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Use case: 12 linked windows with one candidate loaded for each."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "from ipywidgets import Label, Button, HBox, VBox, Layout\n",
+    "\n",
+    "from ginga.misc import Datasrc\n",
+    "from ginga.misc.log import get_logger\n",
+    "from ginga.util import wcsmod\n",
+    "from ginga.util.iohelper import get_fileinfo\n",
+    "\n",
+    "from astrowidgets import ImageWidget"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Reading ASDF in Ginga needs this to be specified early on\n",
+    "wcsmod.use('astropy_ape14')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "logger = get_logger('my viewer', log_stderr=True, log_file=None, level=30)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class LinkedImageWidget(ImageWidget):\n",
+    "    def __init__(self, *args, **kwargs):\n",
+    "        super().__init__(*args, **kwargs)\n",
+    "        num_images = 10  # Max images in cache; should user be able to set this?\n",
+    "        self.datasrc = Datasrc.Datasrc(num_images)  # Cache\n",
+    "        self._other_viewers = []\n",
+    "\n",
+    "    def load_jwst_asdf(self, filename):\n",
+    "        if filename in self.datasrc:\n",
+    "            image = self.datasrc[filename]\n",
+    "        else:\n",
+    "            import asdf\n",
+    "            from ginga.AstroImage import AstroImage\n",
+    "\n",
+    "            image = AstroImage(logger=self.logger)\n",
+    "            image.load_file(filename, data_key='data')\n",
+    "\n",
+    "            self.datasrc[filename] = image\n",
+    "\n",
+    "        self._viewer.set_image(image)\n",
+    "\n",
+    "    def load_fits(self, filename):\n",
+    "        bnch = get_fileinfo(filename)\n",
+    "        \n",
+    "        if filename in self.datasrc:\n",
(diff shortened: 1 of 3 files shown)


mwcraig:
I'll try to take a look at this Wednesday morning.

pllim:
@mwcraig , please let me know if you need access to the data files. Thanks!

hcferguson:
Comments on `linked_wcs_blink.ipynb`.  Nice to see this demo! And that it didn't take huge amounts of code to implement.

- I'm seeing significant lag when I have the image highly zoomed in and I try to pan around. 
- The fact that it flashes up the unzoomed image between blinks is distracting enough that I think we'll have to try to find a fix.
- I find it pretty essential in blinking not to have to move my cursor out of the window. Is it possible to add an event handler for the arrow keys to the image viewer, instead of the `Next Image` button?
- If I pan one image but don't click on the point I want centered, it reverts back to the previous centering as soon as I blink. That was a bit disconcerting.

mwcraig:
@pllim can I please get access to the data files, or a bit more information about them? I have plenty of local FITS files I can use but want to make sure they are similar to yours to evaluate the performance.

hcferguson:
`linked_wcs_*2_windows` both worked fine for me. I see significantly *less* lag on high zoom with these notebooks than in the `linked_wcs_blink.ipynb` case.

pllim:
> significant lag when I have the image highly zoomed

@hcferguson , how high was the zoom? Were you using the images in the example or your own image? Was the image on Central Storage or local disk? I zoomed all the way to 80x and I saw no lag when I pan around.

pllim:
> flashes up the unzoomed image between blinks

Yes, I agree that this should be fixed. There is no such distraction in the [Blink plugin for standalone GUI](https://github.com/ejeschke/ginga/blob/master/ginga/rv/plugins/Blink.py). That plugins seems to take advantage of internal caching of `Channel` object that does not exist on the widget side. @ejeschke, any advise on what is the best way to proceed here?

hcferguson:
@pllim I was using the same images as your example, which I copied to my local disk. The zoom was pretty high, but probably not even as high as 80x before I saw lag.

pllim:
> add an event handler for the arrow keys to the image viewer

I could and it kind of works (see code below). It loads the next image but then it also center to some arbitrary point on the new image that does not align with the previous one, which is also disruptive. If @mwcraig, @eteq, or @ejeschke can point out to me what I am doing wrong here, I would really appreciate it.

I changed this line in the notebook:
```python
d = Event(source=button1, watched_events=['click', 'keydown', 'mouseenter'])
```
to this:
```python
d = Event(source=w1, watched_events=['keydown'])
```

pllim:
> pan one image but don't click on the point I want centered, it reverts back to the previous centering as soon as I blink

Yes, I see this too and I agree that it is undesirable. I think I have an idea on how to fix it. I'll investigate and get back to you. Thanks!

ejeschke:
@pllim, you will want to check on the settings for `autozoom` and `autocenter`. Make sure they are set to 'off'.

pllim:
> pan one image but don't click on the point I want centered, it reverts back to the previous centering as soon as I blink

@hcferguson , I pushed a fix so that this should no longer be an issue. I also make the zoom level auto-sync (for better or worse).

I'll look into the other problems next week.

pllim:
@ejeschke , disabling autozoom/center works. Thanks!

@hcferguson , I think I addressed all [your comments above](https://github.com/astropy/astrowidgets/pull/106#issuecomment-585252768). Please give it another try. p.s. I still cannot reproduce the lag you reported.

hcferguson:
@pllim Took a look at the new version. I saw lag when I first started. Then the lag stopped. Then changed browser windows to write this note. Then went back and it had stopped syncing the windows in position. It still would match zoom when I clicked. To get it to start working again, I had to go back up to the cell starting with `n_widgets = 12` and re-execute all below.

pllim:
Re: https://github.com/astropy/astrowidgets/pull/106#issuecomment-594618559

@hcferguson , again, I cannot reproduce it. I am guessing you are using a Mac laptop with touchpad? Maybe the bindings are behaving differently there. I do not see those reported problems on a Linux desktop. 🤷‍♀

PR #101, opened 2019-12-13
Remove astropy helpers

mwcraig:
This is intended solely to move away from helpers. I haven't done much with the travis matrix yet; I just made the minimal changes needed to make the tests run (maybe)

PR #109, opened 2020-01-30
MNT: Move coveragerc to setup.cfg

pllim:
Follow up of #101


setup.cfg
```diff
@@ -45,6 +45,32 @@ test =
...
+    astrowidgets/_astropy_init*
+    astrowidgets/conftest*
+    astrowidgets/cython_version*
+    astrowidgets/setup_package*
+    astrowidgets/*/setup_package*
+    astrowidgets/*/*/setup_package*
+    astrowidgets/tests/*
+    astrowidgets/*/tests/*
+    astrowidgets/*/*/tests/*
+    astrowidgets/version*
```
astrofrog: You need to repeat all these with a ``*/`` prefix as done here:

https://github.com/astropy/package-template/blob/4e55cdb8a8e0308e4eafde6957edd6b6955a92cc/%7B%7B%20cookiecutter.package_name%20%7D%7D/setup.cfg#L61

Otherwise the coverage won't work well from tox.

pllim: Done.

astrofrog: If you want to do coverage runs from tox you'll also need to add a ``cov:`` entry as in the template, and specify ``--cov-config={toxinidir}/setup.cfg`` - see the template

pllim: Errr... I added it but I didn't write the `tox.ini` for this repo. Maybe @mwcraig can double check if I did it correctly.

mwcraig: Looks like this is correct

PR #110, opened 2020-02-06
[WIP] Example notebook to load large 4D cube

pllim:
**DO NOT MERGE:** Still a work in progress and highly experimental.

This example notebook illustrate how one can load a data cube into a widget and:

1. Have a slider to see desired slice.
2. Have a plot to drill through the slice at a data point you click on.

It also attempts to load a larger-than-memory TESS cube (the experimental part).


example_notebooks/cube_4d.ipynb
```diff
@@ -0,0 +1,260 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# This needs https://github.com/ejeschke/ginga/pull/805\n",
+    "# TODO: Why is it taking so long?\n",
+    "w.load_array(im)"
```
pllim: @ejeschke , `w.load_array(im)` here calls `EnhancedCanvasView.load_data()` on the Dask array. It takes way too long is locking up the computer a little. I am using your PR ejeschke/ginga#805. Is this not the correct way to load a Dask array into the widget viewer?

ejeschke:
>  It takes way too long is locking up the computer a little.

What happens when you try to load a slice from a command line python session. i.e. Just navigate to the slice and then use numpy style slicing on the slice.  Something like 
```python
data_2d = (extract 2D slice)
ht, wd = data_2d.shape
res = data_2d[0:ht:10, 0:wd:10]
```
Does this also take a lot of time?

pllim:
@ejeschke , this is instantaneous. Hope this helps.

```python
>>> import dask.array as da 
>>> from astropy.io import fits 
>>> filename = 'tess-s0001-1-1-cube.fits' 
>>> pf = fits.open(filename, memmap=True) 
>>> im = da.from_array(pf[1].data, name='mytesscube') 
>>> im = im.T  # TESS cube is transposed by ASB 
WARNING: Could not memory map array with mode='readonly', falling back to mode='denywrite', which means that the array will be read-only [astropy.io.fits.file]
>>> obj = (slice(0, 1), slice(0, 1), slice(None), slice(None))                                  
>>> res = im[obj]                                                                               
>>> res                                                                                         
dask.array<getitem, shape=(1, 1, 2136, 2078), dtype=>f4, chunksize=(1, 1, 178, 255), chunktype=numpy.ndarray>
>>> pf.close()
```

pllim:
@ejeschke , I tracked this down to the transpose operation, which is out of scope for Ginga. I have a detail write-up at https://github.com/ejeschke/ginga/pull/805#issuecomment-583495735 . Sorry for the noise!

example_notebooks/cube_4d.ipynb
```diff
@@ -0,0 +1,354 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Loading a fake WCS into widget sometimes gives error.\n",
+    "# Try reloading before giving up.\n",
+    "w.load_nddata(im)"
```
pllim: The first time I call this, I get this error. But when I rerun the cell subsequently, it goes way. @ejeschke , do you know why?

```
2020-02-07 13:05:18,025 | E | Callback.py:172 (_do_callbacks) |
  Error making callback 'cursor-changed': DEC (2048.500000) > 90.0
2020-02-07 13:05:18,026 | E | Callback.py:173 (_do_callbacks) | Traceback:
  File ".../ginga/misc/Callback.py", line 156, in _do_callbacks
    res = method(*cb_args, **cb_kwargs)

  File ".../astrowidgets/core.py", line 214, in _mouse_move_cb
    raDegToString(ra), decDegToString(dec))

  File ".../ginga/util/wcs.py", line 137, in decDegToString
    sign, dec_degree, dec_min, dec_sec = degToDms(dec_deg)

  File ".../ginga/util/wcs.py", line 81, in degToDms
    assert dec <= 90, WCSError("DEC (%f) > 90.0" % (dec))
```

example_notebooks/cube_4d.ipynb
```diff
@@ -0,0 +1,354 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# If plot shows, rerun cell to hide it.\n",
+    "ax = plt.gca();\n",
```
pllim: @mwcraig , do you know any notebook tricks to not show the Matplotlib window in this cell? It ways show on first run, but then disappears in subsequent runs in the same session.

pllim:
@ejeschke , now I encounter a problem where the cursor info bar in

https://github.com/astropy/astrowidgets/blob/604535308e0e33bb08f5a3da3e79be94c5da11ca/astrowidgets/core.py#L192

that is not responsive for this use case. I cannot pin-point to a bottleneck when I run the code piece by piece in an interactive Python session. Any pointers would be greatly appreciated.

example_notebooks/cube_4d.ipynb
ejeschke: That's a WCS check that we probably should look at.  It's just trying to format the coordinates to string values and finding an invalid DEC.  Even though this is a load operation, the formatting happens because of a callback in the viewer when the image changes.  Not sure why it wouldn't occur again if you re-execute the cell, but it might depend on where it thinks the cursor is WRT the window.

ejeschke:
If you add a logging debug message at the beginning of that handler, does it show it being executed?

pllim:
> does it show it being executed?

@ejeschke , thanks for the tip! I inserted some debugging logs, and turns out it was choking at `decDegToString(dec)` conversion.

ejeschke:
> inserted some debugging logs, and turns out it was choking at decDegToString(dec) conversion

I think this could be a general problem in Ginga [noted here](https://github.com/ejeschke/ginga/issues/782), which could use some attention at a hack session sometime.  WCS is tricky that way: when it's broken, you want to know it's broken, but then again you don't want to be *nagged* that its broken. :smiley_cat:

example_notebooks/cube_4d.ipynb
eriniocentric: Hi, I think this works:
plt.ioff()

I've been using a version of this code in a notebook I've made for displaying HETDEX data cubes. I had to switch the axis around.

https://github.com/HETDEX/hetdex_api/blob/master/hetdex_api/cube_widget.py

pllim: The plot doesn't seem to show at the wrong place for me anymore either; perhaps something was fixed upstream. Thanks for letting us know!

Switching axis is probably normal since images taken with different instruments/telescopes have different formats.

pllim:
I am not sure if I'll ever get that "feedback from stakeholders," so @mwcraig , if you think this should go in, then I'll squash the commits. Otherwise, I'll just leave it open. cc @eteq

PR #111, opened 2020-02-10
Ignore RA/Dec string formatting failures

pllim:
While working with #110, I found that invalid WCS causes the cursor info bar to not display at all, when the values raises `AssertionError` in Ginga. I find that it makes more sense to display info bar anyway without RA and Dec, when WCS conversion fails for whatever reason, instead to just not work.

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -214,7 +214,8 @@ class ImageWidget(ipyw.VBox):
                     val += ' (RA: {}, DEC: {})'.format(
                         raDegToString(ra), decDegToString(dec))
                 except Exception:
-                    pass
+                    val += ' (RA, DEC: WCS error)'
+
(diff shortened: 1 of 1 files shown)


mwcraig (approved):
Seems like I can't resist making minor changes this morning. I add a small message indicating an error if setting the RA/Dec fails.

pllim:
For Windows failure, try https://github.com/astropy/astrowidgets/pull/117/commits/7720c6382d37ce9d43ff5d292604101cd6a1a021

mwcraig:
I'll wait to see what works in #117 then rebase....

pllim:
Thanks! 😄

pllim:
Doing a 0.2 release, so that means there won't be any 0.1.2, re-milestoned.

PR #113, opened 2020-02-18
[EXP] CANDELS big mosaic visualization

pllim:
This notebook is an example to visualize a CANDELS mosaic image that is a single-extension FITS with the file size of 4.9 GB. The conclusion of this exercise is that I think Ginga can handle big mosaic natively as-is. I am not sure if this PR needs to be merged; I opened this so others can play around with this notebook on different machines and report back any problems. I did not notice any spikes in memory usage; I think the `memmap` helps a lot here.

p.s. Similar conclusion for Ginga standalone GUI but I did not try using all the plugins on this data, just `Pick` and `Histogram`.

cc @hcferguson

**Example screenshot**

Screenshot from 2020-02-18 11-41-54

### DIFF
--- a/example_notebooks/big_mosaic.ipynb
+++ b/example_notebooks/big_mosaic.ipynb
@@ -1,0 +1,143 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


hcferguson:
Awesome! I'll take a look.

Issue #93, opened 2019-08-20
Refactoring to an abstract interface and concrete implementations


gpdf:
I see that Issue #73 talks about the possibility of a Ginga-free implementation of the ImageWidget API using Matplotlib.  At the time the original API was discussed in 2018 (see, for instance, #10) I had understood that the concept was to define an abstract interface and then, potentially, multiple implementations.

Was this a misunderstanding, or was there a decision to go with a concrete Ginga-based implementation?

Of course, one can always use duck typing to define a family of widgets with compatible APIs, rather than relying on formal subclassing.

LSST has a motivation to do another implementation (based on the [Firefly](https://github.com/Caltech-IPAC/firefly) library and its [Jupyter extension](https://github.com/Caltech-IPAC/jupyter_firefly_extensions).

Would the team be open to a refactoring that realized a design with a more abstract interface?

pllim:
The plan is still the same, but I don't understand your definition of a "more abstract interface". The current API is supposed to be backend-agnostic. If you were to implement a different backend, you do your thing but `load_fits()` will always remain `load_fits`. That is, when you look at `core.py` here, anything *inside* the method can be replaced but the method API should remain the same. Does this make sense?

We have to be careful to not make the interface too complicated. https://xkcd.com/927/

gpdf:
It sounds like you are just suggesting duck typing?

Given the existence of a hypothetical Ginga alternative "FITSWizard", one would create a new package `fitswizardwidgets`, in which I would define an `ImageWidget` class in that namespace, and make its API as close to `astrowidgets.ImageWidget` as possible.  Users would, ideally, only need to change `from astrowidgets import ImageWidget` to `from fitswizardwidgets import ImageWidget` to change from using one to the other?

The down side of this is that there's no simple CI procedure that could be used to discover divergences between the Ginga widget's API and the FITSWizard widget's API, since they are totally independent classes.

What I thought had been envisioned was having a base class that defined the interface, with concrete implementations that implemented it, e.g., `class GingaWidget( ImageWidget )`.

pllim:
I guess abstract base class would work too. Maybe @mwcraig or @eteq can confirm which implementation was the original vision.

mwcraig:
The original intent was indeed to have an abstract API which could be implemented in any number of ways. I had originally thought about bqplot as an alternative backend to  ginga, but as I used the widget thought a matplotlib backend might be helpful too for non-interactive plots.

[...]

PR #121, opened 2021-01-29
EXP: Early exploration with bqplot

### LINKED ISSUE #72: Exploring reimplementation using bqplot
Today I spend some time to explore the possibility of using bqplot for implementation, image from FITS has to be convert to other standard format such as PNG, JPEG for it to work, my question is there better to convert FITS to these format? 

My current way of converting it is using pillow.

pllim:
**Early draft:** More features need to be implemented.

Fix #72 

Would be nice if these get fixed first:

* bqplot/bqplot#284
* bqplot/bqplot#1269
* bqplot/bqplot#1274

Needs:

* glue-viz/bqplot-image-gl#37

After this is merged, need a follow up PR to address the following:

* Address #93 
* Move Ginga out of `core.py` and make it and its dependencies optional.
* Separate example notebooks into sub-folders grouped by backends.

# TODO

- [ ] Maybe `HeatMap` isn't the right thing to use
  - https://github.com/bqplot/bqplot/issues/284#issuecomment-775543692
- [ ] Implement the methods that currently raise `NotImplementedError`
- [ ] Figure out `TODO` items in the in-line comments

### DIFF
--- a/astrowidgets/bqplot.py
+++ b/astrowidgets/bqplot.py
@@ -1,5 +1,7 @@
 """``astrowidgets`` with ``bqplot`` as backend."""
 
+import math
+
 import numpy as np
 
 # Jupyter widgets
@@ -7,6 +9,7 @@ import ipywidgets as ipyw
 
 # bqplot
 from bqplot import Figure, HeatMap, LinearScale, ColorScale, Axis, ColorAxis
+from bqplot_image_gl.interacts import MouseInteraction
 
 __all__ = ['ImageWidget']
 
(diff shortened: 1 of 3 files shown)


pllim:
Got it to load FITS but it ain't pretty using `HeatMap`.

Screenshot 2021-02-08 182432

pllim:
Conclusion: I think this is going too low level. A lot of things would need to be implemented from scratch. I'll try a different approach using a higher level with `glue-jupyter` although I am not sure if that really makes sense because that is basically a Glue backend, not a pure `bqplot` backend.

mwcraig:
I agree that either a glue-based approach or a bqplot + bqplot_image_gl works better here. I'll go ahead and close this but please feel free to reopen if you prefer to keep it open for now.

PR #123, opened 2021-02-18
TST: Fix test header in tox, stricter PEP 8

pllim:
Test header to show properly in tox. Stop excluding files for PEP 8.


pllim:
This should be uncontroversial enough, so merging without a review.

For future reference: `tox` sees only the root level `conftest.py`, while we need `astrowidgets/conftest.py` for the Astropy test runner, so that's why there are two copies. If we decide we don't need to run `astrowidgets.tests()`, we can remove the test runner from `_astropy_init.py` and also remove `astrowidgets/conftest.py`.

PR #124, opened 2021-02-18
DOC: Update RTD config

pllim:
Update RTD config to reflect how we would build the docs locally.


pllim:
Another uncontroversial PR, so merging without review.

PR #125, opened 2021-02-18
TST: Stop ignoring warnings

pllim:
Stop ignoring warnings.

# TODO

- [x] Address warnings seen in logs.
- [x] Turn all warnings into failures and add them as exception as needed.

Issue #122, opened 2021-02-10
API design for choosing between multiple backends


pllim:
Let's say in the future, we have `astrowidgets` API implemented with different visualization tools (backends); e.g., Ginga, glue-jupyter, PyWWT, etc. How do we enable switching backends?

* Do we do this at import time and then have it fixed for a session? Example: `from astrowidgets.ginga import ImageWidget`
* Do we allow user to switch inside an active session? Example: `import astrowidgets; astrowidgets.set_backend('ginga'); from astrowidgets import ImageWidgets`
* Something else I haven't thought of?

Also see: #93

pllim:
With #162  , backend implementation would be completely separate and eventually they won't even be in this repo.

PR #126, opened 2021-02-19
Initial implementation of abstract class

### LINKED ISSUE #93: Refactoring to an abstract interface and concrete implementations

### LINKED ISSUE #127: What is select_points xfailed in test
@mwcraig or @eteq , why is there an xfail test calling select_points ? Is this something obsolete and can be deleted?

### LINKED ISSUE #128: TST: Un-ignore empty marker warning
Follow up of #125 

Need to un-ignore this and handle it appropriately in affected tests with `pytest.warns`.

https://github.com/astropy/astrowidgets/blob/8fed92362886cc56cc71f44a06fe643291f6c33e/setup.cfg#L14

pllim:
Fix #93 

Pre-requisite for #122 

Bonus:

* Fix #127 
* Fix #128 

# TODO

- [x] Rebase after #125 is merged
- [x] Update `docs`
- [x] Update `example_notebooks`
- [x] How to get Ginga `ImageWidget` to inherit docstring from `BaseImageWidget`?
- [x] See if we can get inheritance diagram to show something saner. If not, just do not show.
- [ ] Get feedback from stakeholders
- [ ] After merge, update all my unmerged PRs here


pllim:
@ejeschke , why does the error in https://github.com/astropy/astrowidgets/pull/126/checks?check_run_id=1939249682 (using dev version of Ginga) sounds so familiar. Do you remember how to fix? I don't see that error with dev version of Ginga on my machine.

pllim:
Note to reviewers: I don't expect you to go through all that diffs. This is probably the page you want to focus on: https://astrowidgets--126.org.readthedocs.build/en/126/abstract.html

@mwcraig , also note the API changes reflected in the example notebooks diff.

ejeschke:
@pllim, are any of 'pycairo', 'pillow', 'opencv' or 'aggdraw' installed in the test environment?  One of them is needed for the web backend to function, but I don't think any of them are in `install_requires`. Probably we should add  `pillow` to the requirements because it is used in a lot of places (although technically not required if you are just using, say Qt, with just the ginga widget in a desktop app...)

pllim:
Ah, thanks, @ejeschke ! It is fixed now.

ejeschke:
pycairo will give you the most capability...

pllim:
Re: pycairo -- @ejeschke , for the purpose of CI, that isn't important, but maybe we can discuss this as a separate issue (#130).

mwcraig:
Hoping to take a look at this later today...if not, then first thing wednesday.

pllim:
Thanks, @mwcraig ! Your feedback would be very valuable. I was just playing with finalizing the rendered API appearance and I think I finally got it working.

docs/abstract.rst
```diff
@@ -0,0 +1,258 @@
...
+-----------------
+
+The snippet below is all you should need to make an image widget.
+The widget should be a part of the
+`ipywidgets framework <https://ipywidgets.rtfd.io>`_ so that it can
+be easily integrated with other controls:
+
+.. code-block:: python
+
+    from astrowidgets.somebackend import ImageWidget
```
eteq, line 37: Conceptually I wonder if this is better presented as
```suggestion
    from somevizpackage.astrowidget_backend import ImageWidget
```
? That is, make it clear that external packages can and should write their own backends, rather than it being an import from astrowidgets?

eteq (commented):
Have not managed a proper review of this, but given #142 I figure I should hold off anyway.

Conceptually I like this on its own, but with the caveat of all the points @mwcraig makes in #142. That is, using subclasses like this *might* turn off downstream users who want their astrowidgets backend to *be* the "real" viewer objects. It's easy to imagine a lot of dragons in managing the `__init__`, possible metaclasses, etc.

That said, I think basically everything in the "Understanding BaseImageWidget" section basically is still good even in the #142-style approach!  That is, this captures what the *user* wants to see regardless of whether the backends are implemented as subclasses vs interfaces.

eteq:
In some out-of-band discussion, @mwcraig, @pllim, and I concluded we want to try the `Protocol` approach at least to start with over the abstract/subclass approach implemented here. But the machinery from here for the documentation (i.e. https://astrowidgets--126.org.readthedocs.build/en/126/abstract.html ) and the testing machinery from here should definitely be ported to #142.

pllim:
Superseded by #162

PR #131, opened 2021-02-24
EXP: glupyter implementation on abstract class

pllim:
This builds on #126 (the abstract class) to implement an additional backend using `glue-jupyter` and `bqplot-image-gl`.

Most of the changes are from #126 (they will go away once that PR is merged). The meat of this PR is in:

* `astrowidgets/glupyter.py`
* `example_notebooks/glupyter/`

Known limitations:

* Glue toolbar cannot be displayed because `ipywidgets.VBox` only accept basic widgets as children.
* Glue data collection needs to be cleared before loading new data into the same viewer or APIs like `w.stretch = 'log'` won't work properly (no error but nothing happens on display).

References:

* https://github.com/glue-viz/bqplot-image-gl/blob/master/examples/mouse.ipynb
* https://gist.github.com/astrofrog/7b8ebb3677ff1f5877c6ae82fc8876ec

# TODO

- [ ] Get feedback: Does this make sense? Not with #132 
- [x] See if I can stuff this back into Jdaviz like spacetelescope/jdaviz#428 -- Yes, see spacetelescope/jdaviz#445
- [ ] Implement the rest of functionality for glupyter backend
    - [ ] Marker functionality (set marker properties, click to display, get, remove)
    - [ ] Click to center, controlled by `click_center` flag
    - [ ] Click to drag, controlled by `click_drag` flag
    - [ ] Scroll to pan/zoom, controlled by `scroll_pan` flag -- see if glue-viz/bqplot-image-gl#45 works
    - [ ] Centering image using `center_on()`
    - [ ] Panning using `offset_to()`
    - [ ] Zooming with `zoom()` and `zoom_level`
    - [ ] Autocut using `autocut_options`
    - [ ] Make sure `pixel_offset` usage makes sense
    - [ ] Double check `cursor` positioning still works
- [ ] Add more example notebooks?
- [ ] Add README in `example_notebooks/glupyter/` and polish up notebook narratives
- [ ] Add user-facing docs
- [ ] Add tests
- [ ] Probably need to wait for `glue-jupyter` (and maybe `bqplot-image-gl` too?) release before we can merge this, and also have to pin relevant minversions in `setup.cfg`


astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+        # multiple data into the same widget in the same session.
+        self._app.data_collection.clear()
+
+        self._app.add_datasets(data)
+        viewer.add_data(data)
+        viewer.state.reference_data = data  # Bring it to focus
+        self._link_image_to_cb()
+
+    def center_on(self, point):
+        raise NotImplementedError  # FIXME
```
astrofrog, line 280: You can find and set the current limits with ``viewer.state.x_min``, ``viewer.state.x_max``, ``viewer.state.y_min`` and ``viewer.state.y_max`` so you could use that to compute dx and dy (the field of view) then set e.g. ``viewer.state.x_min = point[0] - dx/2`` and so on if ``point`` is in pixel coordinates

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def offset_to(self, dx, dy, skycoord_offset=False):
```
astrofrog, line 282: Similar to ``center_on``

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+        raise NotImplementedError  # FIXME
+
+    @property
+    def zoom_level(self):
```
astrofrog, line 286: you could get x_min/x_max/y_min/y_max as in center_on and compute the fraction of the image shown and return that as zoom level?

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+        raise NotImplementedError  # FIXME
+
+    @zoom_level.setter
+    def zoom_level(self, value):
```
astrofrog, line 290: Again you can do this by setting x_min/x_max/y_min/y_max

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+        raise NotImplementedError  # FIXME
+
+    # Need this here because we need to overwrite the setter.
+    @property
+    def marker(self):
+        return super().marker
+
+    @marker.setter
+    def marker(self, value):
+        raise NotImplementedError  # FIXME
```
astrofrog, line 300: For all the marker stuff you could do it two ways - keep track of the markers in the astrowidgets layer and use the bqplot ``Scatter`` class to show the markers (you can add/remove markers by setting the ``x`` and ``y`` properties on the scatter object). I can show you an example if needed. Another option would be to actually make the list of markers be a glue data object and add it to the image viewer as a normal glue dataset - the advantage of this is that it shows up in the image viewer layers and you can control the appearance of the markers. There is a glue message you can fire if you want to update the glue dataset to force the viewer to refresh so let me know if you decide to go that route.

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def get_markers_by_name(self, marker_name, x_colname='x', y_colname='y',
+                            skycoord_colname='coord'):
+        raise NotImplementedError  # FIXME
```
astrofrog, line 304: Should be possible if the above is implemented (in both options)

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def add_markers(self, table, x_colname='x', y_colname='y',
+                    skycoord_colname='coord', use_skycoord=False,
+                    marker_name=None):
+        raise NotImplementedError  # FIXME
```
astrofrog, line 309: Also possible

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def remove_markers_by_name(self, marker_name):
+        raise NotImplementedError  # FIXME
```
astrofrog, line 312: Also possible

astrowidgets/glupyter.py
```diff
@@ -0,0 +1,400 @@
...
+    def stretch(self):
+        return self._viewer.state.layers[0].stretch
+
+    @stretch.setter
+    def stretch(self, value):
+        self._viewer.state.layers[0].stretch = value
+
+    @property
+    def autocut_options(self):
+        raise NotImplementedError  # FIXME: Does Glue support this?
```
astrofrog, line 329: I don't understand what this does but you can set the percentile values in glue to automatically determine the levels/cuts in the image viewer. Could you clarify what this should return?

astrofrog (commented):
All the ``NotImplementedError`` things are possible (though for the markers it will require a little bit of code to either keep track of the markers in this class or interface with a glue catalog dataset). Happy to provide examples for various parts of this if needed.

astrowidgets/glupyter.py
pllim: Given that the viewer size and display scale is already known, is it sufficient if I just set `x_min` and `y_min` and then let glupyter take care of recomputing the rest, or is that not a thing?

I am guessing, it takes `x/y_min/max` that is out of range? Otherwise, I cannot center on edges.

astrowidgets/glupyter.py
pllim: I am surprised that zoom info isn't already built into Glue. You have a zoom/pan button after all.

astrowidgets/glupyter.py
pllim: I think the latter makes more sense. Currently in the Ginga implementation, it creates Ginga canvas objects, and then extracts them back from the Ginga canvas, so this is consistent.

What isn't clear to me is how you control the looks of the Glue markers. For example, how do I ask Glue to draw blue X or red circle? How do I only extract markers that are red circle? How do I extract all the markers irrespective of shape/color?

astrowidgets/glupyter.py
pllim: For Ginga, this is what happens:

```python
>>> w.autocut_options
('minmax', 'median', 'histogram', 'stddev', 'zscale')
>>> w.cuts = 'zscale'  # Automatically sets the cut levels according to this algorithm
```

But if this does not make sense for Glue (or other future backends), then maybe I'll take `autocut_options` out of the abstract class and make it a Ginga-only option... 💭

astrowidgets/glupyter.py
pllim: The math isn't quite right. It slowly zooms the image out.

pllim: Okay. I think it is because I am setting only `x_min` and `y_min` and let it auto adjust for the max. Is that not the right way to do this?

mwcraig:
I'm finally taking a look at this batch of PRs and decided to start with this one.  

There are two videos below, one with the implementation in this PR that uses the `BqplotImageView` from `glue-jupyter`. The other uses the mouse demo notebook from https://github.com/glue-viz/bqplot-image-gl so it is displaying a bqplot figure using the `ImageGL` mark. In the glue-baed version there is considerable lag when zooming out or panning.

Is there a way to eliminate the lag in rendering in when zooming out or panning? (maybe an @astrofrog question since the answer is likely in `glue`)?

### glupyter demo

glupyter_demo

### bqplot/ImageGL direct demo

bqplot_image_gl_demo

pllim:
As for the abstract class, I am not sure how to proceed with this PR. I think @astrofrog disagree with some low level implementations here that would warrant refactor even on Ginga's implementation, so there is a high barrier.

For now, we have decided on a "shim" over at Jdaviz. Also see spacetelescope/jdaviz#631 and related PRs over there.

mwcraig:
> As for the abstract class, I am not sure how to proceed with this PR. I think @astrofrog disagree with some low level implementations here that would warrant refactor even on Ginga's implementation, so there is a high barrier.

I might take another shot at it this week -- I have some time through SciPy to devote to it.

I started with this PR because I found it easier to think about the base class after seeing a second implementation of the viewer.

pllim:
@mwcraig , there is also an unresolved comment here: https://github.com/astropy/astrowidgets/issues/93#issuecomment-817941299

pllim:
Most of the work ended up in Jdaviz. This is no longer relevant.

PR #140, opened 2021-07-01
New offset_by with new API

### LINKED ISSUE #139: rename offset_to to offset_byand have it use units
This idea was triggered by looking at the implementation of another astrowidgets-conforming API in spacetelescope/jdaviz#705 :

1. I think the `offset_to` method  is better named `offset_by`.  This [aligns better with astropy.coordinates](https://docs.astropy.org/en/latest/api/astropy.coordinates.offset_by.html) and I think it makes it just a bit clearer that you are giving the *delta* coordinates instead of actual coordinates
2. I think we can dispense with the `skycoord_offset` keyword by instead using the units of `dx`/`dy` - if they are either `pixel` or dimensionless that can be interpreted as pixel offsets, and any other unit can assume `SkyCoord`/other WCS ("other wcs" can be implemented in the future, but I'm  just trying not to accidentally specify ourselves out of it)

My proposal would be to keep in `offset_to` with its current API but have it start raising a deprecation warning (and internally it would call the new `offset_by`)

And yes, `offset_to` was my original suggestion.  But I think it was wrong after seeing it in action!

pllim:
Fix #139 

* New offset_by with new API
* Deprecate offset_to with old API.
* Added tests and updated example notebook.
* Bonus: Fixed a test that was writing file into source dir.

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -6,9 +6,11 @@ import warnings
 
 # THIRD-PARTY
 import numpy as np
+from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.io import fits
 from astropy.table import Table, vstack
+from astropy.utils.decorators import deprecated
 
 # Jupyter widgets
 import ipywidgets as ipyw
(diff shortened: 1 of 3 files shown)


mwcraig (approved):
Thanks, @pllim, merging!

PR #145, opened 2021-07-08
Markers API refactor

### LINKED ISSUE #141: Consider making `marker` private and instead only set it in `add_markers` or `start_marking`
Right now there's a mix of a stateful and a non-stateful idiom in how you set the style of markers.  `self.markers` is settable by the user, but it can also be set in `start_marking`.  `add_markers` also uses `self.markers`, but does *not* have a `markers` keyword.

So I see two ways to fix this:
1. Add a `markers` keyword to `add_markers`
2. Do 1 and *also* make `self.markers` private
3. remove `markers` from `start_marking`

I really don't like 3 because I think it just makes sense to be able to set your marker style when you add the marks.  So that leaves 1 and 2 - 2 is more "pure" (i.e., there's then one way to do it), but has the downside of breaking backwards compatibility (which we may or may not care about). 

Triggered by spacetelescope/jdaviz#699

cc @pllim @mwcraig

### LINKED ISSUE #143: Change marker from a dictionary to a class or `namedtuple`, and change the name from `marker` to `marker_style`
Right now I have to either read the source code or set `iw.marker` to a value I know is bad so that I can be reminded of what the keys/values are for a marker.

If we changed it to a class (or a named tuple) then I could also introspect to get the attributes.

I would also find it clearer if the name were `marker_style`. 

Finally, in an ideal world, we might agree on a few basic shapes and colors that viewers implementing the interface need to support.

pllim:
Fix #141 

Fix #143 

This will propagate to Jdaviz/Imviz too, so once this is merged, I'd rather not change it ever again. Please review carefully. Thanks!

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -31,8 +31,18 @@ ALLOWED_CURSOR_LOCATIONS = ['top', 'bottom', None]
 RESERVED_MARKER_SET_NAMES = ['all']
 
 # Marker style (might be backend specific)
-MarkerStyle = namedtuple('MarkerStyle', ['type', 'color', 'radius'],
-                         defaults=['circle', 'cyan', 20])
+MarkerStyle = namedtuple(
+    'MarkerStyle', ['type', 'color', 'radius', 'linewidth'],
+    defaults=['circle', 'cyan', 20, 1])
+# TODO: Add more examples
+MarkerStyle.__doc__ += """
+
+Marker can be set as follows::
+
+    MarkerStyle(type='circle', color='cyan', radius=20)
+    MarkerStyle(type='cross', color='green', radius=20)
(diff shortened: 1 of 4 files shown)


pllim:
Doh, Actions got deactivated. Closing/re-opening to see if it comes back alive.

mwcraig:
I'll review over the weekend or early next week -- maybe we can try to meet with @eteq at some point early in the week?

pllim:
Has this been superseded by #142 ? I cannot keep track anymore.

mwcraig:
> Has this been superseded by #142 ? I cannot keep track anymore.

🤷‍♂️ Hasn't been decided yet, I think, since we didn't quite get there in the meeting yesterday.

PR #152, opened 2022-02-18
TST: Pin markupsafe in devdeps job

### LINKED ISSUE #151: TST: devdeps broken due to jinja2 and markupsafe conflicts
I have no idea how this is even happening... The answer in pallets/jinja#1593 isn't very helpful (at the time this was posted). All we are doing in that job is installing dev astropy and dev ginga:

https://github.com/astropy/astrowidgets/blob/e54e45cc21a325d554e279b2c9519557ad2e314a/tox.ini#L14-L16

Neither of them should be pulling in any "unsupported jinja". Anyways, if there is not clear resolution upstream in a few days/weeks, we should just pin markupsafe and be done with it. See https://github.com/aws/aws-sam-cli/issues/3661#issuecomment-1044340547

pllim:
Close #151 .

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,6 +12,10 @@ extras = test
 changedir =
   test: .tmp/{envname}
 deps =
+  # TODO: Unpin when unnecessary in the future.
+  # https://github.com/astropy/astrowidgets/issues/151
+  devdeps: markupsafe==2.0.1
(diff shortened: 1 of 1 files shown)


pllim:
huh! didn't even work!

PR #160, opened 2023-03-31
Fix test for stretch

### LINKED ISSUE #159: TST: image.stretch setter fails with Ginga dev
@ejeschke , something changed upstream,  maybe from ejeschke/ginga#1045 ? Can Ginga dev be made backward compatible? Hope you can advise. Thanks!

https://github.com/astropy/astrowidgets/blob/3382c7e2a80eabb76fd802779af4ea85d5ef6837/astrowidgets/core.py#L817

```
_________________________________ test_stretch _________________________________

    def test_stretch():
        image = ImageWidget()
        with pytest.raises(ValueError) as e:
            image.stretch = 'not a valid value'
        assert 'must be one of' in str(e.value)
    
        image.stretch = 'log'
>       assert isinstance(image.stretch, (ColorDistBase))
E       assert False
E        +  where False = isinstance(None, ColorDistBase)
E        +    where None = ImageWidget(...).stretch

astrowidgets/tests/test_api.py:196: AssertionError
```

ejeschke:
- fix #159
- fixes test for "stretch"; use method API to fetch the distribution object rather than accessing instance variable

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -814,7 +814,7 @@ class ImageWidget(ipyw.VBox):
         """
         The image stretching algorithm in use.
         """
-        return self._viewer.rgbmap.dist
+        return self._viewer.rgbmap.get_dist()
 
     # TODO: Possible to use astropy.visualization directly?
     @stretch.setter


pllim (approved):
Looks like it is backwards compatible too. Thanks!

PR #161, opened 2023-03-31
Deprecate use_opencv parameter

ejeschke:
The use_opencv parameter has been deprecated in ginga; opencv will be used automatically if it is installed.

### DIFF
--- a/astrowidgets/core.py
+++ b/astrowidgets/core.py
@@ -48,11 +48,6 @@ class ImageWidget(ipyw.VBox):
     image_width, image_height : int
         Dimension of Jupyter notebook's image widget.
 
-    use_opencv : bool
-        Let Ginga use ``opencv`` to speed up image transformation;
-        e.g., rotation and mosaic. If this is enabled and you
-        do not have ``opencv``, you will get a warning.
-
     pixel_coords_offset : int, optional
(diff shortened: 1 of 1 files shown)


ejeschke:
@pllim, just a suggestion. Not sure how you like to handle deprecation of kwargs.

ejeschke:
Failing test is the one fixed in #160

pllim:
We never released a stable version of this package, so this LGTM. Is there a reason why it is draft? Is there something else you wish to do in this regard? If it is out of draft, I could approve and merge it. Thanks!

ejeschke:
> Is there a reason why it is draft? Is there something else you wish to do in this regard? If it is out of draft, I could approve and merge it.

Just that I was not sure how you liked to handle deprecating kwargs.  This is one of a couple of ways I have used (short of deleting the kwarg), but it has the downside of accepting other kwargs that have no bearing on the constructor.

Another way is to change the default value to None, and then if the passed value is not None, then raise a deprecation error. BUT, someone could still pass the kwarg with None explicitly--thus not realizing that the kwarg is deprecated.

I'll convert from draft--LMK how you like to handle these cases--I'd be interested to know!

pllim (approved):
Since there is no stable release, we don't have to follow normal deprecation best practices. This LGTM. Thanks!

PR #162, opened 2023-05-17
Move interface definition from base class to protocol

mwcraig:
This PR contains the minimal pieces from #142 to change the bottom layer from an abstract base class to a protocol. 

There are some API changes in #142 that are not included in this PR so that we can discuss those separately.


mwcraig:
This also contains the minimal ginga implementation, in part to illustrate how tests would be set up in other implements. See [the ginga-specific test code](https://github.com/astropy/astrowidgets/pull/162/files#diff-08da5e7e3036cc08d4010ce420e8b8b66c802047a337601d410bfa39a92aceee)

pllim:
Thanks! I'll put this on my queue to review.

astrowidgets/interface_definition.py
```diff
@@ -0,0 +1,111 @@
...
+    ALLOWED_CURSOR_LOCATIONS = ('top', 'bottom', None)
+
+    # List of marker names that are for internal use only
+    RESERVED_MARKER_SET_NAMES = ('all', )
+
+    # The methods, grouped loosely by purpose
+
+    # Methods for loading data
+    @abstractmethod
+    def load_fits(self, file):
```
pllim, line 43: I know you say API change is out of scope, but I want to note that we agreed at Astropy Coordination Meeting 2023 to change `load_*` methods to just `load` so Astrowidgets does not dictate what file format each backend choose to support.

https://docs.google.com/document/d/1sBImbw1p4mM3VYWGQJDRbj2N3uqP5Ed_Wfb8BDjJj2o/edit#heading=h.8ah2ugluz5xr

astrowidgets/ginga.py
pllim, line 1: I know you say API change is out of scope, but I want to note that we agreed at Astropy Coordination Meeting 2023 to move backend specific implementations (ginga, bqplot, Firefly, etc) out of `astrowidgets` repo.

https://docs.google.com/document/d/1sBImbw1p4mM3VYWGQJDRbj2N3uqP5Ed_Wfb8BDjJj2o/edit#heading=h.8ah2ugluz5xr

pllim (commented):
Looking at [Astropy Coordination Meeting 2023 notes](https://docs.google.com/document/d/1sBImbw1p4mM3VYWGQJDRbj2N3uqP5Ed_Wfb8BDjJj2o/edit#heading=h.8ah2ugluz5xr), there is a note that says @eteq was going to "split that out into two classes on the train" but I forgot what that was about. Do you remember?

astrowidgets/interface_definition.py
```diff
@@ -0,0 +1,111 @@
...
+    'ALLOWED_CURSOR_LOCATIONS',
+    'RESERVED_MARKER_SET_NAMES'
+]
+
+
+@runtime_checkable
+class ImageViewerInterface(Protocol):
+    # This are attributes, not methods. The type annotations are there
+    # to make sure Protocol knows they are attributes. Python does not
+    # do any checking at all of these types.
```
pllim, line 21: We could enforce it with third-party type checker package but is that going too far?

astrowidgets/interface_definition.py
mwcraig: Thanks -- I added this to the agenda in the running notes.

astrowidgets/interface_definition.py
eteq: Hmm, although it's a bit odd, when I think about it I like the cleanness of the interface not doing any checking but the APITest class doing it. So maybe if we implemented type-checking it could use this information but actually do the type checking when the APITest is invoked by a backend?

eteq (approved):
I like where this is going! I left an inline comment and I have two more general comments below. But I'm approving because all of them might be "lets leave that to a later PR" if you want, @mwcraig .

1. Right now the API methods don't provide any context about what they are supposed to *do*. I'm thinking the interface is now the best place to provide that.  For example, ``load_fits`` might get a docstring that says something like "The UI behavior expected is that the specified FITS file appear in the UI window in addition to being loaded into the application."  Even though we can't guarantee that from an automated testing point of view, we can at least say in the docstring what the API is *supposed* to do from a UI perspective, to make it easier for a user to report variances from a backend to the expected behavior.
2. Maybe we should use type hinting in the methods to also show what each of the methods return? I don't think we need to go full-on with the type-hinting, but as simply a documentation mechanism of the expected return type it might be useful.

astrowidgets/interface_definition.py
pllim: I opened https://github.com/astropy/astrowidgets/issues/163

astrowidgets/ginga.py
pllim: I opened #164 and #165

pllim (approved):
In principle, I think Protocol is a great idea and we should push this forward. Let's get this in. Thanks!

Matt, in case if you want to address Erik comments, I won't merge now. Feel free to merge when you think it is ready.

astrowidgets/ginga.py
```diff
@@ -55,6 +49,11 @@ class ImageWidget(ipyw.VBox):
         correct value to use.*
 
     """
+    # Allowed locations for cursor display
+    ALLOWED_CURSOR_LOCATIONS = ('top', 'bottom', None)
+
+    # List of marker names that are for internal use only
+    RESERVED_MARKER_SET_NAMES = ('all', )
```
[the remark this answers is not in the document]
mwcraig: Yeah, they should only be in one place. It feels a little inelegant to do it as constants. Will update.

mwcraig:
> Right now the API methods don't provide any context about what they are supposed to do. I'm thinking the interface is now the best place to provide that. For example, load_fits might get a docstring that says something like "The UI behavior expected is that the specified FITS file appear in the UI window in addition to being loaded into the application." Even though we can't guarantee that from an automated testing point of view, we can at least say in the docstring what the API is supposed to do from a UI perspective, to make it easier for a user to report variances from a backend to the expected behavior.

Going to implement that in this PR because I know from experience I will never come back to it...

astrowidgets/interface_definition.py
mwcraig: Agreed -- and the way we have set things up I don't think we can.

astrowidgets/interface_definition.py
```diff
@@ -0,0 +1,255 @@
...
+    click_drag: bool
+    scroll_pan: bool
+    image_width: int
+    image_height: int
+    zoom_level: float
+    is_marking: bool
+    stretch_options: tuple
+    autocut_options: tuple
+    cursor: str
+    marker: Any
```
mwcraig, line 32: These were omitted the first time through:

    zoom_level: float
    is_marking: bool
    stretch_options: tuple
    autocut_options: tuple
    cursor: str

Issue #167, opened 2023-06-19
Add more types to the protocol


mwcraig:
As a follow-up to #162 add type hints to the methods/attributes of the API interface. See [this comment](https://github.com/astropy/astrowidgets/pull/162#pullrequestreview-1483554902).

mwcraig:
Closed by #168

PR #168, opened 2023-06-26
Add type hints to methods of interface, including arguments

mwcraig:
Pretty sure I did this right, but this is my first genuine attempt at type hinting, so 🤷‍♂️ 

Related to #167

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -9,7 +9,7 @@ sphinx:
 
 # Set the version of Python and requirements required to build your docs
 python:
-  version: 3.8
+  version: 3.10
   system_packages: false
   install:
     - method: pip
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,8 +8,12 @@ sphinx:
   fail_on_warning: true
 
 # Set the version of Python and requirements required to build your docs
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
+
 python:
-  version: 3.10
   system_packages: false
   install:
     - method: pip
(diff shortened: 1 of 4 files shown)


mwcraig:
Test failure here is because I'm using the `Union` notation introduced in `Python 3.10`, i.e. I'm doing `str | None` as opposed to `Union[str, None]` or `Optional[str`. 

Are we ok with making 3.10 the minimum version?

pllim:
> Are we ok with making 3.10 the minimum version?

Fine by me. I think all this means is downstream needs that version to check against Protocol, but doesn't mean they themselves need to support that minversion, right?

mwcraig:
I think that is correct. If we split the API out into its own package then users never need to install the API package; it could simply be a testing dependency for the downstream packages.

pllim:
In that case, bumping to Python 3.10 is not controversial. :shipit:

mwcraig:
I think this is ready for merge -- I've updated the minimum python version 3.10 so that we can use the `|` operator in type hints.

.readthedocs.yml
```diff
@@ -9,7 +9,7 @@ sphinx:
 
 # Set the version of Python and requirements required to build your docs
 python:
-  version: 3.8
+  version: 3.10
```
pllim, line 12: You have to use the new build syntax. See https://docs.readthedocs.io/en/stable/config-file/v2.html

mwcraig:
All tests are passing now -- thanks for the RTD fix, @pllim!

pllim (approved):
Looks uncontroversial enough. Thanks!

PR #170, opened 2023-07-06
Add overwrite argument to save method

mwcraig:
I was discussing astrowidgets with a colleague who pointed out that, by default, `astrowidgets` silently overwrites files when it saves. This PR adds an `overwrite` argument to the `save` method and implements it in the ginga viewer. I'm hoping this is uncontroversial enough that it can be approveed/merged without a synchronous discussion 😀

### DIFF
--- a/astrowidgets/ginga.py
+++ b/astrowidgets/ginga.py
@@ -2,6 +2,7 @@
 
 # STDLIB
 import functools
+from pathlib import Path
 import warnings
 
 # THIRD-PARTY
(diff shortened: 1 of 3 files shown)


astrowidgets/ginga.py
```diff
@@ -973,12 +974,23 @@ def scroll_pan(self, value):
...
             bindmap.map_event(None, (), 'pa_pan', 'zoom')
 
-    def save(self, filename):
+    def save(self, filename, overwrite=False):
         """
         Save out the current image view to given PNG filename.
+
+        Parameters
+        ----------
+        filename : str
```
pllim, line 983: Should this be str or Path too?

pllim (approved):
LGTM thanks! Have a minor question about the Ginga implementation but no objection on adding this keyword.

When this is merged, I probably have to go back to Jdaviz and check the compliance.

astrowidgets/ginga.py
mwcraig: Heck, probably should be both. Will fix in a separate PR since I was a little too quick on the merge button this morning

PR #171, opened 2023-07-07
Better document type options

mwcraig:
This addresses [the comment in #170](https://github.com/astropy/astrowidgets/pull/170#discussion_r1254556606) about adding more detail in the docstring about typing. I did it two different ways, not sure which is the correct one to use....


pllim (commented):
Which one works better with Protocol?

mwcraig:
I think (based on some minimal googling) that `os.PathLike` is the way to go.

PR #173, opened 2023-07-27
Update the interface and ginga implementation to make some changes of the marker API

mwcraig:


astrowidgets/tests/widget_api_test.py
```diff
@@ -243,10 +243,48 @@ def test_add_markers(self):
         with pytest.raises(ValueError, match="No markers named 'test2'"):
             self.image.get_markers(marker_name='test2')
 
+    def test_get_markers_accepts_list_of_names(self):
+        # Check that the get_markers method accepts a list of marker names
+        # and returns a table with all the markers from all the named sets.
+        rng = np.random.default_rng(1234)
+        data = rng.integers(0, 100, (5, 2))
```
pllim, line 250: Is randomness really necessary here?

astrowidgets/tests/widget_api_test.py
```diff
@@ -243,10 +243,48 @@ def test_add_markers(self):
...
+        self.image.add_markers(tab, marker_name='test2')
+
+        # No guarantee markers will come back in the same order, so sort them.
+        t1 = self.image.get_markers(marker_name=['test1', 'test2'])
+        # Sort before comparing
+        t1.sort('x')
+        expected = vstack([tab, tab], join_type='exact')
+        expected.sort('x')
+        assert np.all(t1['x'] == expected['x'])
+        assert (t1['y'] == expected['y']).all()
```
pllim, line 262:
```suggestion
        np.testing.assert_array_equal(t1['x'], expected['x'])
        np.testing.assert_array_equal(t1['y'], expected['y'])
```

astrowidgets/tests/widget_api_test.py
```diff
@@ -243,10 +243,48 @@ def test_add_markers(self):
...
 
+    def test_remove_markers_name_all(self):
+        rng = np.random.default_rng(1234)
+        data = rng.integers(0, 100, (5, 2))
+        tab = Table(data=data, names=['x', 'y'])
+        self.image.add_markers(tab, marker_name='test1')
+        self.image.add_markers(tab, marker_name='test2')
+
+        self.image.remove_markers(marker_name='all')
+        assert self.image.get_markers(marker_name='all') is None
```
pllim, line 276: If we return empty table instead, this test will need to be updated, no?

astrowidgets/tests/widget_api_test.py
```diff
@@ -243,10 +243,48 @@ def test_add_markers(self):
...
+
+    def test_remove_marker_accepts_list(self):
+        rng = np.random.default_rng(1234)
+        data = rng.integers(0, 100, (5, 2))
+        tab = Table(data=data, names=['x', 'y'])
+        self.image.add_markers(tab, marker_name='test1')
+        self.image.add_markers(tab, marker_name='test2')
+
+        self.image.remove_markers(marker_name=['test1', 'test2'])
+        assert self.image.get_markers(marker_name='all') is None
```
pllim, line 286: If we return empty table instead, this test will need to be updated, no?

pllim (commented):
As discussed today, I will review this ignoring potential refactoring we need for regions support.

Not a fan of the recursion in Ginga implementation but that works for now. We might have to refactor again anyway.

I don't think you need `np.random`. The test data is small enough and the context is irrelevant, so something like `np.arange(10).reshape((5, 2))` would work just as well.

astrowidgets/tests/widget_api_test.py
mwcraig: Yes, that is correct

astrowidgets/tests/widget_api_test.py
pllim: Is the empty table thingy part of this PR or separate PR?

astrowidgets/tests/widget_api_test.py
mwcraig: No, that is just what GitHub Copilot inserted 😬

mwcraig:
Thanks, @pllim -- I'll hold off on changing the return values until after today's meeting.

astrowidgets/tests/widget_api_test.py
mwcraig: Since we are meeting later today it may as well be part of this PR. If the next meeting was pretty far off I'd be inclined to keep it spearate.

astrowidgets/tests/widget_api_test.py
```diff
@@ -138,12 +144,13 @@ def test_marking_operations(self):
         assert not self.image.click_drag
         assert not self.image.scroll_pan
 
-        # Regression test for GitHub Issue 97:
-        # Marker name with no markers should give warning.
-        with pytest.warns(UserWarning, match='is empty') as warning_lines:
+        # Make sure no warning is issued when trying to retrieve markers
+        # with a name that does not exist.
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
```
pllim, line 150: Hmm, won't setting pytest to turn all warnings into exceptions take care of this?

astrowidgets/tests/widget_api_test.py
```diff
@@ -234,19 +241,58 @@ def test_add_markers(self):
...
+        expected.sort('x')
+        np.testing.assert_array_equal(t1['x'], expected['x'])
+        np.testing.assert_array_equal(t1['y'], expected['y'])
 
     def test_remove_markers(self):
         with pytest.raises(ValueError, match='arf'):
             self.image.remove_markers(marker_name='arf')
 
+    def test_remove_markers_name_all(self):
+        rng = np.random.default_rng(1234)
```
pllim, line 276: There are remaining unnecessary random number generators that should be replaced with static arrays.

astrowidgets/tests/widget_api_test.py
```diff
@@ -234,19 +241,58 @@ def test_add_markers(self):
...
+        data = rng.integers(0, 100, (5, 2))
+        tab = Table(data=data, names=['x', 'y'])
+        self.image.add_markers(tab, marker_name='test1')
+        self.image.add_markers(tab, marker_name='test2')
+
+        self.image.remove_markers(marker_name='all')
+        self._check_marker_table_return_properties(self.image.get_markers(marker_name='all'))
+
+    def test_remove_marker_accepts_list(self):
+        rng = np.random.default_rng(1234)
```
pllim, line 286: And here.

mwcraig:
Lights are green, so this is ready for another look

pllim (commented):
In the light of https://github.com/spacetelescope/jdaviz/pull/2410 , I find that the API will be significantly different if we were to go with Regions instead of Table.

So, do you want to merge this anyway and refactor again later, or wait for that discussion first?

mwcraig:
Right, forgot about that. Let's wait until after the next meeting.

PR #192, opened 2025-07-11
Add plotly as a backend

marthacryan:


mwcraig:
Thanks, @marthacryan -- I got bogged down in other things at sprints, but will come back to this in early August.

PR #205, opened 2026-07-04
Update ginga viewer to match AIDA interface

mwcraig:
This pull request updates the ginga viewer to use the astro-image-display-api. The commit history is a little noisy -- I was using AI to help update the code and wanted small commits I could more easily review -- but I'm happy to squash before merging.

This also contains the broken import that appears in #204, so once this is merged a rebase of #204 should fix that one.


pllim:
I guess you and the AI better hash things out first before I approve?

mwcraig:
Good call -- turns out that the bqplot viewer that I thought was done is definitely not done...lesson being do more local tests in a browser before pushing.

astrowidgets/ginga.py
```diff
@@ -7,31 +7,129 @@
...
+    Because every astropy stretch maps the unit interval onto itself, the
+    stretch evaluated on ginga's 0..1 ramp *is* the lookup table ginga needs,
+    so this adapter reproduces any astropy stretch exactly.
+    """
+
+    def __init__(self, hashsize, stretch, colorlen=None):
+        self.stretch = stretch
+        super().__init__(hashsize, colorlen=colorlen)
+
+    def calc_hash(self):
```
mwcraig, line 57: Every method needs a numpy-style docstring

astrowidgets/ginga.py
```diff
@@ -7,31 +7,129 @@
...
+    Ginga's parametrized distributions are reparametrizations of astropy's
+    stretches, so where ginga has the family we configure its native class:
+    the hyperbolic families match exactly, while log/power match the curve
+    shape to within a quantization level (ginga normalizes by ``log(a)``/``a``
+    where astropy uses ``log(a + 1)``/``a - 1``). Stretches ginga lacks --
+    including ``PowerStretch`` (x**a), which is a different family than ginga's
+    ``'power'`` (astropy's ``PowerDistStretch``) -- are handled exactly by
+    `_AstropyStretchDist`.
+    """
+    if isinstance(stretch, AsinhStretch):
```
mwcraig: I think this would be more readable as a `match`

astrowidgets/ginga.py
```diff
@@ -7,31 +7,129 @@
...
+        return ColorDist.SquaredDist(hashsize)
+    if isinstance(stretch, SqrtStretch):
+        return ColorDist.SqrtDist(hashsize)
+    if (isinstance(stretch, LinearStretch)
+            and stretch.slope == 1 and stretch.intercept == 0):
+        return ColorDist.LinearDist(hashsize)
+    return _AstropyStretchDist(hashsize, stretch)
+
+
+def docs_from_super_if_missing(cls):
```
mwcraig, line 162: Question: should this live in astro-image-display-api?

astrowidgets/ginga.py
```diff
@@ -43,61 +141,57 @@ class ImageWidget(ipyw.VBox):
...
-    DEFAULT_INTERACTIVE_MARKER_NAME: str = ImageViewerInterface.DEFAULT_INTERACTIVE_MARKER_NAME
-
-    def __init__(self, logger=None, image_width=500, image_height=500,
-                 pixel_coords_offset=0, **kwargs):
-        super().__init__()
+    DEFAULT_INTERACTIVE_MARKER_NAME: str = 'interactive'
 
+    def __init__(self, *args, logger=None, image_width=500, image_height=500,
+                 **kwargs):
         if 'use_opencv' in kwargs:
```
mwcraig: This can be dropped -- that deprecated thing is long gone from ginga.

astrowidgets/ginga.py
```diff
@@ -223,682 +304,342 @@ def _mouse_move_cb(self, viewer, button, data_x, data_y):
...
+        self._viewer.set_image(self._build_ginga_image(data))
+
+        # Apply the stored viewport, cuts, stretch and colormap to the freshly
+        # displayed image.
+        self._apply_viewport_to_ginga(image_label)
+        self._apply_cuts_to_ginga(image_label)
+        self._apply_stretch_to_ginga(image_label)
+        self._apply_colormap_to_ginga(image_label)
+
+    def _is_displayed_label(self, image_label):
```
mwcraig, line 423: Do these docstrings numpy-style (Parameters, Returns, Notes sections)

astrowidgets/ginga.py
```diff
@@ -223,682 +304,342 @@ def _mouse_move_cb(self, viewer, button, data_x, data_y):
...
+    # ------------------------------------------------------------------
+    # Cuts / stretch / colormap
+    # ------------------------------------------------------------------
+    def set_cuts(self, value, image_label=None, **kwargs):
+        super().set_cuts(value, image_label=image_label, **kwargs)
+        # Changing the cuts only affects the color mapping, so leave the
+        # current viewport (zoom/pan) untouched.
+        self._apply_cuts_to_ginga(image_label)
+
+    def _apply_cuts_to_ginga(self, image_label):
```
mwcraig, line 501: Even private methods get proper docstrings!

astrowidgets/ginga.py
```diff
@@ -43,61 +140,53 @@ class ImageWidget(ipyw.VBox):
...
 
+    def __init__(self, *args, logger=None, image_width=500, image_height=500,
+                 **kwargs):
         if 'use_opencv' in kwargs:
+            # Pop it so it is not forwarded to VBox, which would raise a
+            # TraitError for an unknown trait.
+            kwargs.pop('use_opencv')
             warnings.warn("use_opencv kwarg has been deprecated--"
                           "opencv will be used if it is installed",
                           DeprecationWarning)
```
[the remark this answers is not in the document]
mwcraig: This turns out not to be correct on the current stack, though the concern is understandable. With ipywidgets 8.1.7 / traitlets 5.14.3, an unrecognized keyword argument does not raise a `TraitError` — traitlets emits a `DeprecationWarning` and construction succeeds:

```
DeprecationWarning: Passing unrecognized arguments to super(ImageWidget).__init__(pixel_coords_offset=1).
```

`test_pixel_offset_removed` passes as written, and its comment notes this is deliberate: unlike `use_opencv` (which gets a targeted pop-and-warn), `pixel_coords_offset` intentionally falls through to traitlets' generic deprecation path. The caveat is that traitlets says this will become an error in a future release — when that happens the test will fail and flag that a targeted pop is needed.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
```diff
@@ -223,682 +302,309 @@ def _mouse_move_cb(self, viewer, button, data_x, data_y):
...
-        self._viewer.set_pan(pan_x + dx_val, pan_y + dy_val, coord=dx_coord)
-
-    @property
-    def zoom_level(self):
+        Build a ginga drawing-object factory from a catalog style dict.
         """
-        Zoom level:
+        shape = style.get('shape', 'circle')
+        color = style.get('color', 'red')
+        size = style.get('size', 5)
```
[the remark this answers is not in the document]
mwcraig: Confirmed — this is a real gap. `_make_marker` only reads `shape`, `color`, and `size` from the style dict, so the `linewidth` that the AIDA logic stores (and that the notebook passes) is silently dropped. Ginga's canvas objects do accept it (verified: `dc.Circle(x=1, y=1, radius=5, color='red', linewidth=3)` constructs fine and honors the value), so this is a small fix in `_make_marker`. It will be addressed in a follow-up commit.

*This comment was written by Claude at my direction.*

example_notebooks/ginga_widget.ipynb
```diff
@@ -200,10 +231,10 @@
     "# example image.\n",
     "ra_str = '01h13m23.193s'\n",
     "dec_str = '+00d12m32.19s'\n",
-    "frame = 'icrs'\n",
+    "frame = 'galactic'\n",
```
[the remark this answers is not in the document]
mwcraig: Confirmed, and it is worse than just naming: `SkyCoord('01h13m23.193s', '+00d12m32.19s', frame='galactic')` parses as (l, b) = (18.35°, 0.21°), which is ~102° on the sky from the ICRS position those strings are meant to describe, so the cell centers the viewer far off the example image. I also verified the downstream effect: after setting a galactic center, `get_viewport()['center']` comes back as a Galactic `SkyCoord` with no `.ra` attribute, so the later offset cell's `hasattr(current_center, 'ra')` check silently skips. The cell should use `frame='icrs'`; will fix in a follow-up commit.

*This comment was written by Claude at my direction.*

example_notebooks/ginga_widget.ipynb
```diff
@@ -534,17 +603,27 @@
...
     "def show_circles(n):\n",
-    "    \"\"\"Show and hide circles.\"\"\"\n",
-    "    w.reset_markers()\n",
+    "    \"\"\"Show subset of stars using catalog API.\"\"\"\n",
+    "    # Remove existing catalog\n",
+    "    try:\n",
+    "        w.remove_catalog(catalog_label='stars')\n",
+    "    except ValueError as exc:\n",
+    "        raise exc  # Catalog doesn't exist\n",
+    "        \n",
```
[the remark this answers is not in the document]
mwcraig: Confirmed. `remove_catalog` does raise `ValueError` for a missing label (verified: `ValueError: Catalog label stars not found.`), and catching it only to immediately `raise exc` is a no-op — running the slider cell before the catalog-loading cell would still crash. The except body should be `pass` (with the "catalog doesn't exist" comment) to make out-of-order execution graceful. Will fix in a follow-up commit.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
mwcraig: Agreed. `calc_hash` and `get_dist_pct` override hooks from ginga's `ColorDistBase`, and `calc_hash` in particular has the non-obvious side effect of setting `self.hash` rather than returning a value — worth spelling out in a numpy-style docstring. Will add these in a follow-up commit.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
mwcraig: Agreed, this works and reads well. The package's `python_requires >=3.10` covers `match`, class patterns give the same isinstance semantics with first-match-wins ordering (so the SquaredStretch-before-PowerStretch subtlety is preserved by keeping the same case order), and the `LinearStretch` condition maps cleanly to an attribute pattern: `case LinearStretch(slope=1, intercept=0):`. The only mild counterpoint is that each branch returns immediately, so the current chain is already flat — the gain is aesthetic rather than structural. Will convert in a follow-up commit.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
mwcraig: Good candidate for upstreaming, yes. The strongest evidence is that this exact helper is already duplicated verbatim in `astrowidgets/bqplot.py` on this branch — every AIDA backend will want it. The one generalization needed is that it currently hard-codes `ImageViewerLogic` as the docstring source; an upstream version could take the interface class as an argument or walk the decorated class's MRO. Worth opening an issue on astro-image-display-api and dropping both local copies once it lands there.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
mwcraig: Agreed on the substance: ginga 5.3.0 (the version in the dev environment) contains no reference to `use_opencv` anywhere in the installed package, and the astrowidgets deprecation shipped in v0.4.0 (March 2023), so it has had a multi-release deprecation period. Two side effects to handle when dropping it: (1) `test_use_opencv_kwarg_warns_not_errors` needs to be removed or updated; (2) anyone still passing `use_opencv=` will then get traitlets' generic "unrecognized arguments" `DeprecationWarning` instead of the targeted message — still non-fatal on current traitlets, same path `pixel_coords_offset` takes. Will drop it in a follow-up commit.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
mwcraig: Agreed — `_is_displayed_label` and its neighbors have prose docstrings but not the numpy-style structure. Will restructure them with Parameters/Returns/Notes sections in a follow-up commit.

*This comment was written by Claude at my direction.*

astrowidgets/ginga.py
mwcraig: Fair — `_apply_cuts_to_ginga` has inline comments but no docstring, and the sibling `_apply_stretch_to_ginga` / `_apply_colormap_to_ginga` helpers are in the same state. Will add proper docstrings to all of the `_apply_*_to_ginga` helpers in a follow-up commit.

*This comment was written by Claude at my direction.*

example_notebooks/ginga_widget.ipynb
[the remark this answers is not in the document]
mwcraig: Um, no you should not change this. The image's WCS is in galactic coordinates. Instead of setting this back to ICRS and having the image display incorectly, fix the offset cell by adding a small offset and remove the attribute check.

astrowidgets/ginga.py
mwcraig: Make sure you do a proper docstring for every method except those that inherit from `ImageViewerLogic`.

mwcraig:
@pllim -- if you are interested this is ready for review. Honestly, it is such a big change that the diff is not very meaningful. In any event, I'll leave it open a couple days but merge buy the weekend if you don't have time

pllim:
Yeah I am not sure if I can get to `+1,435 -899` in a timely manner. Is there a specific part I should focus on?

mwcraig:
If you have a chance to try it out locally this morning that would be great -- if not, I have tried it out myself locally.

PR #207, opened 2026-07-06
Fix several issues with the bqplot viewer

mwcraig:
It turns out the bqplot had a number of issues:

+ The default colormap was terrible; it is now "Greys_r".
+ The default cuts were terrible; the ones shipped in AIDA's image viewer logic would have been better but it wasn't being used. I ended up defaulting to one a little different than the AIDA one (higher black level). Note that this is *not* something the AIDA API requires -- default cuts can be whatever a viewer wants -- but the bare-bones image logic in AIDA shipped a interval that would at least allow you to see the image.
+ Changing the cuts or stretch reset the color map and whichever of the cuts or stretch was not being set. The tests for this should at some point be upstreamed to the set of tests that AIDA ships for viewers.
+ Changing which image was displayed was....painful to watch. A slow-mo sequence of resetting the color map, cuts, and stretch back to the defaults on the current image, changing its zoom to fit in the viewer, and then displaying the new image. That is all hidden now by holding of on syncing to the front end until all widgets are ready for that.
+ Changing the image reset the colormap, cuts and stretch. I think users will not expect that so now the current settings are kept except for viewport, which resets (though that can be suppressed with an argument to load_image).
+ An image label of `None` -- which is the label used when the user has not specified one -- was not handled correctly.

This PR fixes all of those things. The number of commits is large to ensure that each change is bite-sized and easy to review.

### DIFF
--- a/astrowidgets/bqplot.py
+++ b/astrowidgets/bqplot.py
@@ -395,6 +395,9 @@ class ImageWidget(ipw.VBox, ImageViewerLogic):
         # Guards re-entrancy while we programmatically update the viewport.
         self._updating_viewport = False
 
+        # While True, _refresh_display does nothing; set by _defer_refresh.
+        self._refresh_deferred = False
+
         # Provide an Output widget to which prints can be directed for
         # debugging.
         self._print_out = ipw.Output()
@@ -561,12 +564,25 @@ class ImageWidget(ipw.VBox, ImageViewerLogic):
         Recompute the displayed array from the cuts and stretch stored
         for the image and send it to the viewer.
         """
-        if self._data is None:
+        if self._data is None or self._refresh_deferred:
             return
 
         self._send_data(cuts=self.get_cuts(image_label=image_label),
                         stretch=self.get_stretch(image_label=image_label),
                         reset_view=reset_view)
+
+    @contextmanager
+    def _defer_refresh(self):
+        """
+        Make _refresh_display a no-op inside the block so that several
+        settings can be stored with a single recomputation of the
+        displayed array. The caller refreshes after the block.
+        """
+        self._refresh_deferred = True
+        try:
+            yield
+        finally:
+            self._refresh_deferred = False
 
     @property
     def _current_image_label(self):
@@ -593,21 +609,47 @@ class ImageWidget(ipw.VBox, ImageViewerLogic):
 
     # The methods, grouped loosely by purpose
     def load_image(self, image, image_label=None, **kwargs):
+        # A newly loaded image is always displayed with the current cuts,
+        # stretch and colormap; only the viewport resets. Capture the
+        # current settings before the API layer replaces them with its own
+        # defaults during the load.
+        if self._data is not None:
+            # A new image: carry forward from the displayed image.
+            settings_label = self._current_image_label
+        else:
+            # Nothing has been displayed yet.
+            settings_label = None
+
+        if settings_label is not None:
+            current_cuts = self.get_cuts(image_label=settings_label)
+            current_stretch = self.get_stretch(image_label=settings_label)
+            current_colormap = self.get_colormap(image_label=settings_label)
+        else:
+            current_cuts = self._default_cuts
+            # Leave the stretch the API layer stores on load (linear).
+            current_stretch = self._default_stretch
+            current_colormap = self._default_colormap
+
         # Hold the sync so the scale changes from the viewport
         # initialization in the API layer arrive at the front end in the
         # same batch as the new image data, avoiding flicker from
         # intermediate redraws.
         with self._astro_im._hold_all_sync():
-            super().load_image(image, image_label=image_label, **kwargs)
-            data = self.get_image(image_label=image_label)
+            # Store the settings for the new image so the get_* methods
+            # report what is displayed. Defer the refresh each setter
+            # triggers; one refresh at the end displays the image.
+            with self._defer_refresh():
+                super().load_image(image, image_label=image_label, **kwargs)
 
-            self._data = data.data if isinstance(data, NDData) else data
+                self.set_cuts(current_cuts, image_label=image_label)
+                if current_stretch is not None:
+                    self.set_stretch(current_stretch, image_label=image_label)
+                self.set_colormap(current_colormap, image_label=image_label)
+
+                data = self.get_image(image_label=image_label)
+                self._data = data.data if isinstance(data, NDData) else data
+
             self._refresh_display(image_label=image_label, reset_view=True)
-
-            # The API layer does not store a colormap on load, so record the
-            # default so that get_colormap reports what is displayed.
-            if self.get_colormap(image_label=image_label) is None:
-                self.set_colormap(self._default_colormap, image_label=image_label)
 
     # Saving contents of the view and accessing the view
     def save(self, filename, overwrite=False, **kwargs):
--- a/astrowidgets/bqplot.py
+++ b/astrowidgets/bqplot.py
@@ -392,7 +392,10 @@ class ImageWidget(ipw.VBox, ImageViewerLogic):
         self._default_cuts = apviz.AsymmetricPercentileInterval(30, 96)
         self._default_stretch = None
         self._default_colormap = 'Greys_r'
-        self._astro_im.set_color(bqcolors(self._default_colormap))
+        # Store the default through the API layer (which keeps settings
+        # for the None label even before a load) so that get_colormap
+        # reports it; set_colormap also applies it to the front end.
+        self.set_colormap(self._default_colormap)
 
         self._data = None
         self._wcs = None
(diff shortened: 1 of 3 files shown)


astrowidgets/bqplot.py
```diff
@@ -364,8 +387,12 @@ def __init__(self, *args, display_width=500, display_aspect_ratio=1):
 
         self._astro_im = _AstroImage(display_width=display_width,
                                      viewer_aspect_ratio=display_aspect_ratio)
-        self._default_cuts = apviz.MinMaxInterval()
+        # Cut out the sky background at the bottom and clip only the
+        # brightest pixels at the top.
+        self._default_cuts = apviz.AsymmetricPercentileInterval(30, 96)
         self._default_stretch = None
+        self._default_colormap = 'Greys_r'
+        self._astro_im.set_color(bqcolors(self._default_colormap))
```
mwcraig: Can this be set with the API set_colormap?

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -30,6 +32,256 @@ def test_save(self, tmp_path):
...
+
+        cuts = self.image.get_cuts()
+        assert isinstance(cuts, apviz.AsymmetricPercentileInterval)
+        assert cuts.lower_percentile == 30
+        assert cuts.upper_percentile == 96
+
+        displayed = np.asarray(self.image._astro_im._image.image)
+        np.testing.assert_allclose(displayed, cuts(arr))
+
+    def test_set_stretch_keeps_stored_cuts(self):
```
mwcraig, line 220: How does this check whether stored outs are preserved, since it uses the default widget cuts?

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -30,6 +32,256 @@ def test_save(self, tmp_path):
...
+
+        self.image.load_image(arr, image_label='second')
+
+        assert spy_image.call_count <= 1
+        assert spy_x.call_count <= 1
+        assert spy_y.call_count <= 1
+
+        # Batching must not change the end state.
+        displayed = np.asarray(image_mark.image)
+        expected = self.image.get_cuts(image_label='second')(arr)
```
mwcraig: Would it strengthen this test to set the cuts to something other than the default?

astrowidgets/bqplot.py
```diff
@@ -550,26 +604,68 @@ def set_stretch(self, value, image_label=None, **kwargs):
...
-        data = self.get_image(image_label=image_label)
+        # A newly loaded image is always displayed with the current cuts,
+        # stretch and colormap; only the viewport resets. Capture the
+        # current settings before the API layer replaces them with its own
+        # defaults during the load.
+        if image_label is not None and image_label in self._images:
+            # Re-loading an existing image: its own settings are current.
+            settings_label = image_label
+            have_current = True
+        elif self._data is not None:
```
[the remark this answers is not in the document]
mwcraig: The scenario described here can't actually occur, because the API layer's label resolution prevents `image_label=None` from ever addressing the stored `None` entry once a user-defined label exists. In `ImageViewerLogic._resolve_image_label`:

- **0 user labels:** `None` stays `None`, so the unlabeled entry's own settings are what get carried forward — which *is* the preserving behavior.
- **Exactly 1 user label:** `None` resolves to that label — for `load_image`, `set_cuts`, `get_cuts`, all of it. So in the scenario described (a `None`-labeled image exists, a labeled image is displayed), `load_image(..., image_label=None)` replaces the *labeled* image and stores the carried-forward settings under that label. The `None` entry and its settings are never touched.
- **2+ user labels:** `_resolve_image_label` raises `ValueError`, so the load fails before anything is stored.

Verified empirically on this branch:

```python
w = ImageWidget()
w.load_image(arr)                                   # unlabeled
w.set_cuts(apviz.ManualInterval(10, 20))
w.load_image(arr + 1, image_label='x')
w.set_cuts(apviz.ManualInterval(30, 40), image_label='x')
w.load_image(arr + 2, image_label=None)             # "reload the unlabeled image"

# keys: [None, 'x']; the new data landed under 'x', not None
# _images[None].cuts is still the original ManualInterval(10, 20) object
```

It's also impossible to *create* a `None`-labeled entry after a labeled image exists (an unlabeled `load_image` resolves to the existing label), so `None` can only be the current image when it's the only image — and then the `self._data is not None` branch carries forward its own stored settings, exactly the preservation this comment asks for.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -30,6 +32,256 @@ def test_save(self, tmp_path):
...
+        displayed = np.asarray(self.image._astro_im._image.image)
+        np.testing.assert_allclose(displayed, first_stretch(first_cuts(new_arr)))
+
+    def test_load_image_keeps_settings_without_labels(self):
+        # The carry-forward of cuts, stretch and colormap must work in the
+        # common interactive case where no image_label is ever passed, so
+        # every image and its settings live under the API's default (None)
+        # label. Loading a second image must keep the settings currently in
+        # effect, not fall back to the widget defaults.
+        rng = np.random.default_rng(seed=42)
```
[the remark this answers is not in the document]
mwcraig: Step (3) of the proposed test isn't expressible in this API: once a labeled image exists, `load_image` with no label is resolved by `ImageViewerLogic._resolve_image_label` to the single existing user label, so it replaces the *labeled* image — there is no way to reload the unlabeled one. The original unlabeled settings therefore *are* preserved after that sequence (verified: `_images[None].cuts` is still the original object), but trivially so, because the reload never touches the `None` entry. Asserting "not carried forward from the labeled one" would test a code path that can't be reached.

That said, a regression test pinning down the *actual* mixed labeled/unlabeled behavior would be worthwhile: load unlabeled + customize, load labeled + customize, then check that a third unlabeled load (1) replaces the labeled image while keeping that label's settings, (2) leaves the `None` entry's settings untouched, and (3) raises `ValueError` once two user labels exist. Happy to add that to this branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/bqplot.py
mwcraig: Yes, it can. I expected the API layer to refuse before any image is loaded, but `ImageViewerLogic` keeps a `None`-label placeholder entry from construction, so `self.set_colormap(self._default_colormap)` works here and produces byte-identical front-end colors (the override calls `_astro_im.set_color(bqcolors(cmap_name, reverse=False))`, and `reverse=False` is also `bqcolors`'s default, so the two calls are equivalent).

Verified on this branch:

```python
w = ImageWidget()          # no image loaded
w.set_colormap('viridis')  # no error; placeholder stores it
w.get_colormap()           # 'viridis' — reported even before the first load
```

Going through the API is a small improvement beyond style: `get_colormap()` then reports the default before the first load instead of `None`. First-load behavior is unchanged, since `load_image` re-applies the default through `set_colormap` anyway. I can swap line 395 to `self.set_colormap(self._default_colormap)` when you want.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_widget_api_bqplot.py
mwcraig: You're right — it doesn't, fully. After a bare `load_image` the stored cuts *are* `self._default_cuts` (load stores that exact object), so this assertion cannot distinguish "the refresh used the stored cuts" from "the refresh fell back to the widget defaults": both produce `stretch(default_cuts(arr))`. It still catches cruder regressions (refreshing with no cuts at all, or not refreshing after `set_stretch`), but not the specific fallback its name and comment claim.

To make it discriminating, set non-default cuts first:

```python
self.image.load_image(arr)
cuts = apviz.AsymmetricPercentileInterval(5, 90)
self.image.set_cuts(cuts)
stretch = apviz.LogStretch()
self.image.set_stretch(stretch)

displayed = np.asarray(self.image._astro_im._image.image)
np.testing.assert_allclose(displayed, stretch(cuts(arr)))
```

(The sibling `test_set_cuts_keeps_stored_stretch` doesn't have this weakness — the `LogStretch` it sets differs from the default of no stretch.) Happy to make this change on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_widget_api_bqplot.py
mwcraig: Modestly, yes. As written, the cuts carried forward to `'second'` are the widget defaults (nothing custom was set on `'first'`), so `expected = get_cuts('second')(arr)` would also pass under a batched-refresh bug that fell back to `self._default_cuts`. Setting custom cuts on `'first'` before the second load would make this end-state check able to catch that.

Two caveats on how much it buys: the real teeth of this test are the `send_state` call-count assertions, and the discriminating settings-carry-forward end state is already pinned down by `test_load_image_keeps_settings_carried_forward` (which uses custom cuts, stretch, and colormap and checks the displayed array). So it closes no coverage gap, but it does make this test's sanity check meaningful on its own — a two-line change, happy to add it.

— _Written by Claude at @mwcraig's direction._

pllim:
Just merge when you ready. Thanks!

PR #210, opened 2026-07-15
Fix bqplot named marker rendering

### LINKED ISSUE #208: bqplot backend: "diamond", "plus" and "crosshair" markers silently render as nothing; marker size ignored
### Description

In the bqplot backend, several marker shapes that `astro-image-display-api` documents as supported are silently not drawn at all, and the marker `size` in `catalog_style` is ignored.

Versions: astrowidgets 0.5.1, bqplot 0.12.45, bqplot-image-gl 1.8.0, astro-image-display-api (current release).

### 1. `diamond`, `plus` and `crosshair` markers render as nothing, with no warning

The AIDA `set_catalog_style` docs say:

> The following shapes are supported: "circle", "square", "crosshair", "plus", "diamond".

`plot_named_markers` in `astrowidgets/bqplot.py` passes the shape name straight to a bqplot `ScatterGL` mark. The Python-side `marker` trait accepts all 12 bqplot marker names, but the ScatterGL **WebGL frontend** only implements 6 of them. From the fragment shader in the bqplot labextension bundle:

```glsl
#define FAST_CIRCLE 1
#define FAST_SQUARE 2
#define FAST_ARROW 3
#define FAST_CROSS 4
#define FAST_TRIANGLE_UP 5
#define FAST_TRIANGLE_DOWN 6
```

For any other shape name, `FAST_DRAW` is never defined, the shader's `inner_shape`/`outer_shape` stay `0.0`, and the marker is drawn fully transparent. No error or warning is raised anywhere — the marker just doesn't appear. Three of the five AIDA-documented shapes (`diamond`, `plus`, `crosshair`) fall in this hole.

Reproducer:

```python
import numpy as np
from astropy.table import Table
from astrowidgets.bqplot import ImageWidget

iw = ImageWidget()
iw.load_image(np.random.normal(size=(100, 100)))
iw.load_catalog(
    Table({"x": [30.0], "y": [40.0]}),
    catalog_label="cat",
    catalog_style={"shape": "diamond", "color": "red", "size": 20},
)
# Nothing appears. With "square" instead of "diamond" a marker appears.
```

We hit this in stellarphot ([stellarphot#584](https://github.com/feder-observatory/stellarphot/pull/584)): APASS comparison stars styled with `"diamond"` and exclusion markers styled with `"plus"` were invisible, while `"square"` markers in the same viewer rendered fine.

**Recommended fix** (*edited 2026-07-09 after field-testing; see the follow-up comment below for details*): construct a plain SVG `Scatter` instead of `ScatterGL` in `plot_named_markers`.

- `ScatterGL` subclasses `Scatter` and overrides only `_view_name`/`_model_name` (bqplot 0.12.45), so it is trait-for-trait identical — a one-word change at the single construction site.
- The SVG frontend draws all 12 marker shapes, fixing this bug outright.
- Field-tested in stellarphot via the monkeypatch `astrowidgets.bqplot.ScatterGL = bqplot.Scatter`: diamonds and plusses render correctly, SVG marks compose fine over the `ImageGL` layer, and pan/zoom stays smooth at realistic catalog sizes.
- In bqplot 0.13 the GL marks move out to the separate [`bqplot-gl`](https://pypi.org/project/bqplot-gl/) package — that's what broke the import in #196, closed with an upper pin. Staying on `ScatterGL` under 0.13 would mean a new dependency and an import change, and would buy nothing for this bug: bqplot-gl's `scatter-fragment.glsl` has the identical six-shape `FAST_*` list. Switching to `Scatter` lets the `bqplot<0.13` pin be lifted with no new dependency.

The trade-off is rendering performance for very large catalogs (SVG creates a DOM node per point and slows down somewhere in the thousands of points). If that matters, an opt-in GL path could be kept, translating the shape names the shader can't draw (e.g. `plus`/`crosshair` → GL `cross`, which renders as a plus sign) and warning when a shape can't be honored.

### 2. Marker `size` is ignored

`plot_named_markers` hardcodes `default_size=100` and never uses its `size` argument:

```python
sc = ScatterGL(scales=scale_dict,
               x=x, y=y,
               colors=[color],
               default_size=100,
               marker=shape,
               fill=False)
```

so the `size` in `catalog_style` (passed as `size=size**2` from `set_catalog_style`) has no effect. The fix is to pass `default_size=size` — the value arriving from `set_catalog_style` is already squared, i.e. already in the px² units `default_size` expects.

### 3. (minor) Confusing traittypes warning on every catalog load

`plot_named_markers` passes the catalog table's `Column` objects to the `ScatterGL` `x`/`y` traits. `np.asarray` copies an ndarray subclass even when the dtype matches, so traittypes emits the baffling

```
UserWarning: Given trait value dtype "float64" does not match required type "float64". A coerced copy has been created.
```

on every catalog load (`np.dtype(None).name` is `"float64"`, hence the self-contradictory message). Calling `np.asarray(...)` on the coordinates before handing them to the mark would avoid it.

---

*This issue was researched and written by Claude (Anthropic's AI assistant) at the direction of @mwcraig, who reviewed it before posting. Edited 2026-07-09, also by Claude at @mwcraig's direction, to replace the list of candidate fixes with the field-tested recommendation.*

sapunyangkut:
## Summary

- render named bqplot markers with `Scatter` so supported non-circle shapes are visible
- pass the requested marker size through to `default_size`
- add a regression test for the mark type, shape, and size

## Root cause

`plot_named_markers` always constructed `ScatterGL`, whose frontend does not render every marker shape accepted by its Python trait, and it hardcoded `default_size=100` instead of using the method's `size` argument.

## Validation

- regression test failed on the unmodified base with `ScatterGL` and `default_size=100`, then passed with this change
- `python -m pytest -p no:cacheprovider --pyargs astrowidgets docs -q --tb=short` (`70 passed, 2 skipped`)
- `python -m flake8 astrowidgets --count` (`0`)
- `git diff --check`
- local secrets-only scan (`No secrets detected`)

The automated regression test verifies the Python mark class and trait state rather than browser pixels. The issue's maintainer has separately reported real-application field testing of this approach.

Closes #208

This pull request was prepared with OpenAI Codex assistance. I reviewed the change, diff, and test results.


mwcraig:
@sapunyangkut -- thanks for the pull request! I switched it to "ready for review" because from my initial look it is ready.

mwcraig (changes requested):
Thanks for opening the PR! The code that is here is fine, but can you please also fix "Confusing traittypes warning on every catalog load"?

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -18,6 +18,17 @@ def test_instance():
     assert isinstance(image, ImageViewerInterface)
 
 
+def test_named_marker_uses_scatter_shape_and_size():
+    image = ImageWidget()
+    image._astro_im.plot_named_markers(
```
mwcraig: Can you please call the ImageWidget public API to check this? It is image.set_catalog_style(...).

One note: the actual marker size that goes into `_astro_im.plot_named_markers` is squared by `set_catalog_style`

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -18,6 +18,17 @@ def test_instance():
...
+        np.array([1.0]), np.array([2.0]), "test",
+        size=49, shape="diamond")
```
mwcraig: Please test the full range of shapes in addition to diamond: `circle`, `square`, `crosshair`, `plus`, `diamond`

mwcraig (changes requested):
Apologies for not catching these in my first review. There is one other thing from #208 that is unaddressed: the size is set incorrectly by `load_catalog` -- let me know whether you want to fix that here too.

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -45,6 +48,35 @@ def user_callback(interaction, event_data, buffers):
     assert calls == [click_event]
 
 
+@pytest.mark.parametrize(
+    "shape", ["circle", "square", "crosshair", "plus", "diamond"]
+)
+def test_catalog_markers_use_scatter_shape_size_and_arrays(shape):
+    image = ImageWidget()
+
+    with warnings.catch_warnings():
```
mwcraig: This warning handling isn't needed because tests are always run in a way that promotes warnings to errors. Now that this warning has been removed from the exceptions in pyproject.toml it will raise an error if it reappears.

astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -45,6 +48,35 @@ def user_callback(interaction, event_data, buffers):
...
+        warnings.filterwarnings("error", message="Given trait value dtype")
+        image.load_catalog(
+            Table({"x": [1.0], "y": [2.0]}),
+            catalog_label="test",
+            catalog_style={"color": "red", "shape": "circle", "size": 6},
```
mwcraig: Please go ahead let the parameter shape change here too.
```suggestion
            catalog_style={"color": "red", "shape": shape, "size": 6},
```

mwcraig (changes requested):
Thanks again for the changes -- this should be my last round of change requests. Please make your changes as an additional commit instead of force pushing a new commit -- that makes it easier for me to review 😀

mwcraig (approved):
Thanks again for the pull request!

PR #215, opened 2026-07-17
Slim the bqplot backend to AIDA rendering hooks

### LINKED ISSUE #213: bqplot backend: set_catalog_style and remove_catalog use the unresolved catalog_label as the mark id
`ImageWidget.set_catalog_style` passes the caller's `catalog_label` straight to `plot_named_markers` as `str(catalog_label)` without resolving it first.

With one catalog loaded as `"test"`, calling `image.set_catalog_style(size=7)` (no label) correctly stores the style under `"test"` but plots a second, orphan `Scatter` under mark id `"None"` — the old markers keep the old style and a duplicate marker set appears in the figure:

```python
from astropy.table import Table
from astrowidgets.bqplot import ImageWidget

im = ImageWidget()
im.load_catalog(Table({'x': [1.0], 'y': [2.0]}), catalog_label='test')
print(list(im._astro_im._scatter_marks))  # ['test']
im.set_catalog_style(size=7)
print(list(im._astro_im._scatter_marks))  # ['test', 'None'] -- orphan mark
print(im.get_catalog_style())             # style correctly stored under 'test'
```

`remove_catalog(None)` has the mirror problem: the catalog data is removed, but `remove_named_markers("None")` raises `ValueError` because the scatter actually lives under `"test"`.

The fix is to resolve the label before using it as the mark id, e.g. take it from `get_catalog_style()["catalog_label"]` (the API's `get_catalog_style` returns the resolved label) rather than using the raw argument.

This could also be folded into the planned shared draw-catalog refactor across backends.

mwcraig:
This slims the bqplot backend down to the rendering-hook interface added to `ImageViewerLogic` in astro-image-display-api 0.3.0 (the template-method refactor).

Now that #205 and #214 have merged, this branch has been rebased onto `main` and contains only this PR's three commits; the diff is just `astrowidgets/bqplot.py` and its test file. The `astro-image-display-api>=0.3.0` pin (released) landed with #214.

Closes #213.

### What changed

- Deleted the public-method wrappers (`load_image`, `set_cuts`, `set_stretch`, `set_colormap`, `set_viewport`, `get_viewport`, `load_catalog`, `set_catalog_style`, `remove_catalog`), including `load_image`'s carry-forward block, which now lives upstream. The display glue moved into the hooks: `_render_image`, `_apply_cuts`, `_apply_stretch`, `_apply_colormap`, `_apply_viewport`, `_draw_catalog`, `_remove_catalog_marks`, built on the existing internals (`_refresh_display`, `_send_data`, `plot_named_markers`/`remove_named_markers`).
- `_batch_update` composes the two existing batching context managers via an `ExitStack`: the front-end sync of the image mark and the scales is held for the whole block (one state message per widget, image mark flushing first, so no flicker or refit flash), and the refreshes triggered by the `_apply_*` hooks are deferred and collapsed into a single recomputation of the displayed array that runs while the sync is still held. The existing batching regression tests pass unchanged.
- Because the hooks receive resolved labels, the catalog scatter marks are always stored under the resolved catalog label — fixes #213 (unlabeled `set_catalog_style` plotted an orphan mark under the id `"None"`; unlabeled `remove_catalog` raised). Regression tests added.
- Fixed a latent `__init__` bug: the default colormap was stored through `set_colormap` before any image was loaded, which now raises. The default is applied directly to the front end instead, and `_render_image` stores the widget defaults (Greys_r, 30–96 percentile cuts) for the first image displayed, i.e. a load with nothing to carry forward.
- Deleted the local `docs_from_super_if_missing` in favor of the upstream decorator, and the `_current_image_label` property; the GUI pan/zoom observer now targets the displayed image's label explicitly.
- Marker interaction helpers and the rest of the bqplot-specific public API are untouched.
- Widget tests that encoded the old None-label and reload semantics are updated to the released AIDA 0.3.0 behavior: an unlabeled load targets the single shared default label, so repeated unlabeled loads replace each other while never touching an explicitly labeled image (see the review discussion below).

`test_widget_api_bqplot.py` passes in full (97 passed, 2 skipped) and the ginga suite is unaffected (110 passed); lint (`ruff check astrowidgets`) is clean.

Generated by Claude Code at Matt Craig's direction.

### DIFF
--- a/astrowidgets/bqplot.py
+++ b/astrowidgets/bqplot.py
@@ -555,10 +555,6 @@ class ImageWidget(ipw.VBox, ImageViewerLogic):
         finally:
             self._refresh_deferred = False
 
-    @property
-    def viewer(self):
-        return self._astro_im
-
     # ------------------------------------------------------------------
     # Rendering hooks
     #
--- a/astrowidgets/bqplot.py
+++ b/astrowidgets/bqplot.py
@@ -15,7 +15,10 @@ import ipywidgets as ipw
 from matplotlib import pyplot
 from matplotlib.colors import to_hex
 
-from astro_image_display_api import ImageViewerLogic, docs_from_super_if_missing
+from astro_image_display_api.image_viewer_logic import (
+    ImageViewerLogic,
+    docs_from_image_viewer_logic_if_missing,
+)
 
 
 class _AstroImage(ipw.VBox):
@@ -350,7 +353,7 @@ def bqcolors(colormap, reverse=False):
 
 
 # The inheritance order below matters -- VBox needs to come first
-@docs_from_super_if_missing
+@docs_from_image_viewer_logic_if_missing
 class ImageWidget(ipw.VBox, ImageViewerLogic):
     def __init__(self, *args, display_width=500, display_aspect_ratio=1):
         super().__init__(*args)
--- a/astrowidgets/bqplot.py
+++ b/astrowidgets/bqplot.py
@@ -1,4 +1,4 @@
-from contextlib import contextmanager
+from contextlib import ExitStack, contextmanager
 from pathlib import Path
 
 import numpy as np
@@ -15,34 +15,7 @@ import ipywidgets as ipw
 from matplotlib import pyplot
 from matplotlib.colors import to_hex
 
-from astro_image_display_api.image_viewer_logic import ImageViewerLogic
-
-
-def docs_from_super_if_missing(cls):
-    """
-    Decorator to copy the docstrings from the interface methods to the
-    methods in the class.
-    """
-    for name, method in cls.__dict__.items():
-        if not name.startswith("_"):
-            if method.__doc__:
-                continue
-            # Not sure why this fails, but it does.
-            # method.__doc__ = inspect.getdoc(method)
-            interface_method = getattr(ImageViewerLogic, name, None)
-
-            if interface_method:
-                method.__doc__ = interface_method.__doc__
-            # print(f"{method} {method.__doc__} {inspect.getdoc(method)} {interface_method.__doc__=}")
-            # supers = cls.__mro__
-            # for a_super in supers:
-            #     print(f"{name=} {a_super=}")
-            #     interface_method = getattr(a_super, name, None)
-            #     if interface_method:
-            #         print(f"{name=} {a_super=}")
-            #         method.__doc__ = interface_method.__doc__
-            #         break
-    return cls
+from astro_image_display_api import ImageViewerLogic, docs_from_super_if_missing
 
 
 class _AstroImage(ipw.VBox):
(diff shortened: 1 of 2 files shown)


astrowidgets/tests/test_widget_api_bqplot.py
```diff
@@ -213,59 +243,55 @@ def test_load_image_keeps_settings_without_labels(self):
...
         np.testing.assert_allclose(displayed, stretch(cuts(arr2)))
 
-    def test_load_image_without_label_targets_single_labeled_image(self):
-        # An image label of None means "the caller did not specify a
-        # label". Once a labeled image exists, the API layer resolves None
-        # to that label, so an unlabeled load replaces the labeled image
-        # (keeping its settings) and never touches an image stored under
-        # the None label. With two or more labels the target is ambiguous
-        # and the load raises.
+    def test_load_image_without_label_never_replaces(self):
```
mwcraig: A loaded image without label is supposed to *always* replace

mwcraig: Partly right, and the test's wording is stale either way. In the released AIDA 0.3.0, `load_image` with no label always resolves to the single shared default label (`_resolve_label` with `allow_new=True` returns `DEFAULT_LABEL`). So:

- Repeated unlabeled loads **do** always replace the previous unlabeled image (keeping its cuts/stretch/colormap, per the `load_image` docstring) — verified: after two unlabeled loads only `_internal_default_label` exists.
- An unlabeled load **never** replaces a *labeled* image — after `load_image(arr, image_label='labeled')` followed by an unlabeled load, both `labeled` and the default label are loaded, with the default one displayed.

The "generated labels" comments and the `never_replaces` name here were written against pre-release AIDA behavior, where each unlabeled load got a unique generated label. The assertions still pass against 0.3.0 (the labeled image is indeed untouched), but the name and comments describe the wrong mechanism. Follow-up commit will rename this to something like `test_load_image_without_label_targets_default_label`, fix the comments, and add an assertion that a second unlabeled load replaces the first.

*(Reply written by Claude Code at Matt's direction.)*

mwcraig: Addressed in 5088320: renamed to `test_load_image_without_label_targets_default_label`, reworded the stale generated-label comments (including the catalog test), and added an assertion that a second unlabeled load replaces the first in place. *(Written by Claude Code at Matt's direction.)*

astrowidgets/bqplot.py
```diff
@@ -572,78 +556,211 @@ def _defer_refresh(self):
             self._refresh_deferred = False
 
     @property
-    def _current_image_label(self):
+    def viewer(self):
```
mwcraig: Why was `viewer` added?

mwcraig: For parity with the ginga backend, which has long exposed a public `viewer` property ("The underlying ginga viewer object", returning the `EnhancedCanvasView`) as the escape hatch to the raw rendering object. Here it returns `self._astro_im`, the internal bqplot figure wrapper, in the same spirit.

Nothing actually depends on it — no code, test, doc, or notebook uses `.viewer` on the bqplot widget. So it's optional: it could be dropped (defensible, since unlike ginga's viewer, `_AstroImage` is an internal class rather than an object from another library), or kept with a docstring matching ginga's. Happy to do either.

*(Written by Claude Code at Matt's direction.)*

mwcraig: Yes, drop it

mwcraig: Dropped in 988929b. *(Written by Claude Code at Matt's direction.)*

PR #218, opened 2026-07-18
Switching to ruff linter

anzelpwj:
Updating to a more modern linter. Made some small changes to fix things. Using the old 79 per line character limit. E402 is for where imports are in ipynb cells (irrelevant for just checking in `astrowidgets`, but I wanted to test `ruff check .`), and F403 lets us do the star import in `docs/conf.py` (again, not in `astrowidgets` but still trying to run on all).


mwcraig (approved):
Thanks @anzelpwj!

PR #222, opened 2026-07-19
Factor the cursor-info readout into a module shared by both backends

### LINKED ISSUE #203: ginga viewer displays coordinates as "RA/Dec" even if the coordinates are actually galactic
The sample image downloaded for use in the example notebooks has a WCS that is in galactic coordinates. The ginga viewer displays those galactic coordinates but labels them RE/Dec instead of ℓ/b.

mwcraig:
The new `astrowidgets/cursor_info.py` holds a pure `format_cursor_text()` function and a `CursorInfoMixin` that owns the HTML readout widget, the `cursor` location property, and the data/WCS lookup through the AIDA logic layer, so the ginga and bqplot backends now produce identical readouts from a one-line forward of their native mouse-move events.

Along the way this fixes several readout bugs and rough edges:

- Restores the `cursor` property (`'top'`/`'bottom'`/`None`) on the bqplot backend, which was dropped in the AIDA slim-down (#215).
- Fixes the bqplot readout calling `pixel_to_world(y, x)` with swapped arguments, which displayed the sky position of the transposed pixel.
- Adds a `sky_coordinate_format` property selecting decimal degrees (the new default for both backends) or sexagesimal for the RA/Dec display; changing it re-renders the readout immediately at the last cursor position instead of waiting for the next mouse move.
- The sky position is always converted to ICRS whatever the native frame of the image WCS, and the readout is tagged `(ICRS)` to say so. (The pre-refactor ginga readout showed the native world coordinates — galactic l/b for the Spitzer example image — mislabelled as RA/Dec, with the longitude formatted in hours.)
- The X/Y and decimal RA/Dec fields are fixed width and the readout renders in a `<pre>` block, so the line no longer jitters side-to-side as the cursor moves.
- Uses floor-based rounding for the value lookup so cursor positions slightly outside the image no longer read pixel 0.

The shared behavior is covered by a new test module that runs the same assertions against both backends, including a test that both produce byte-identical readout strings for the same image and cursor position.

Fixes #203 

This PR was generated by Claude Code at Matt's direction.
Claude-Session: https://claude.ai/code/session_013M3uRgBJF5BwAVXhNhakTo

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,8 +3,27 @@
 
 TBD -- First stable release.
 
+New Features
+------------
+
+- The cursor-information readout (X/Y, RA/Dec, pixel value) is now shared
+  between the ginga and bqplot backends via
+  ``astrowidgets.cursor_info.CursorInfoMixin``. The ``cursor`` property
+  ('top', 'bottom', or ``None``), which had been dropped from the bqplot
+  backend, is available again on both backends, and a new
+  ``sky_coordinate_format`` property selects between decimal degrees (the
+  new default for both backends) and sexagesimal for the RA/Dec display.
+  As part of this change the ginga readout switched from sexagesimal to
+  decimal degrees by default; set ``sky_coordinate_format`` to
+  ``'sexagesimal'`` for the previous behavior.
+
 Bug Fixes
 ---------
+
+- Fixed the RA/Dec shown in the bqplot ``ImageWidget`` cursor readout,
+  which transposed the x and y pixel coordinates when converting to sky
+  coordinates. Also fixed both backends reading pixel 0's value for
+  cursor positions slightly outside the image on the negative side.
 
 - Fixed the bqplot ``ImageWidget`` so that mouse clicks no longer raise
   ``AttributeError`` and no longer block user-registered ``on_msg``
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,14 +8,12 @@ New Features
 
 - The cursor-information readout (X/Y, RA/Dec, pixel value) is now shared
   between the ginga and bqplot backends via
-  ``astrowidgets.cursor_info.CursorInfoMixin``. The ``cursor`` property
-  ('top', 'bottom', or ``None``), which had been dropped from the bqplot
-  backend, is available again on both backends, and a new
-  ``sky_coordinate_format`` property selects between decimal degrees (the
-  new default for both backends) and sexagesimal for the RA/Dec display.
-  As part of this change the ginga readout switched from sexagesimal to
-  decimal degrees by default; set ``sky_coordinate_format`` to
-  ``'sexagesimal'`` for the previous behavior.
+  ``astrowidgets.cursor_info.CursorInfoMixin``. Both backends have a
+  ``cursor`` property (``'top'``, ``'bottom'``, or ``None``) and a new
+  ``sky_coordinate_format`` property that selects decimal degrees (the
+  default) or sexagesimal for the RA/Dec display. The ginga readout now
+  defaults to decimal degrees; set ``sky_coordinate_format`` to
+  ``'sexagesimal'`` for the previous display. [#222]
 
 Bug Fixes
 ---------
--- a/astrowidgets/cursor_info.py
+++ b/astrowidgets/cursor_info.py
@@ -63,6 +63,9 @@ def format_cursor_text(x, y, data=None, wcs=None, sky_format='degrees'):
         raise ValueError(f'Invalid value {sky_format!r} for sky_format. '
                          f'Valid values are: {ALLOWED_SKY_COORDINATE_FORMATS}')
 
+    if data is not None:
+        data = np.asarray(data)
+
     # floor rather than int() so that positions just outside the image on
     # the negative side do not alias onto pixel 0.
     x_index = int(np.floor(x + 0.5))
@@ -93,6 +96,11 @@ def format_cursor_text(x, y, data=None, wcs=None, sky_format='degrees'):
                 dec = f'{sky.dec.deg:<+8.4f}'
             segments.append(f'RA: {ra} Dec: {dec} (ICRS)')
         except Exception:
+            # Deliberately broad: the WCS is user-supplied and only
+            # duck-typed to the APE-14 interface, so the exceptions it can
+            # raise are open-ended, and this runs on every mouse move,
+            # where an escaped exception would spam the log or vanish
+            # instead of surfacing cleanly. Degrade to a visible error.
             segments.append('RA/Dec: WCS error')
 
     segments.append(value)
@@ -107,21 +115,24 @@ class CursorInfoMixin:
     `~astro_image_display_api.image_viewer_logic.ImageViewerLogic`, whose
     per-label image storage supplies the data and WCS for the readout.
     The mixin defines no ``__init__``; backends call `_init_cursor_info`
-    from their own ``__init__`` and forward their native mouse-move
-    events to `_update_cursor_text`.
+    from their own ``__init__``, place the returned readout widget in
+    their own ``children``, and forward their native mouse-move events
+    to `_update_cursor_text`.
     """
     ALLOWED_CURSOR_LOCATIONS = ALLOWED_CURSOR_LOCATIONS
     ALLOWED_SKY_COORDINATE_FORMATS = ALLOWED_SKY_COORDINATE_FORMATS
 
-    def _init_cursor_info(self, image_widget):
+    def _init_cursor_info(self):
         """
-        Create the readout widget and lay it out below ``image_widget``.
+        Create and return the readout widget. The backend adds it to its
+        own ``children``, below the image widget; the `cursor` property
+        flips the box to ``column-reverse`` to show it on top.
         """
         self._cursor_readout = ipw.HTML('Coordinates show up here')
         self._sky_coordinate_format = 'degrees'
         self._last_cursor_position = None
-        self.children = [image_widget, self._cursor_readout]
         self.cursor = 'bottom'
+        return self._cursor_readout
 
     @property
     def cursor(self):
@@ -133,6 +144,18 @@ class CursorInfoMixin:
 
     @cursor.setter
     def cursor(self, val):
+        """
+        Set the readout location.
+
+        Parameters
+        ----------
+        val : str or None
+            One of `ALLOWED_CURSOR_LOCATIONS`: ``'top'`` or ``'bottom'``
+            place the readout above or below the image by flipping the
+            box's flex direction; ``None`` hides it. The readout is
+            re-rendered at the last cursor position. Anything else
+            raises `ValueError`.
+        """
         if val is None:
             self._cursor_readout.layout.visibility = 'hidden'
             self._cursor_readout.layout.display = 'none'
@@ -159,6 +182,18 @@ class CursorInfoMixin:
 
     @sky_coordinate_format.setter
     def sky_coordinate_format(self, val):
+        """
+        Set the RA/Dec display format.
+
+        Parameters
+        ----------
+        val : str
+            One of `ALLOWED_SKY_COORDINATE_FORMATS`: ``'degrees'`` or
+            ``'sexagesimal'``. The readout is immediately re-rendered
+            at the last cursor position, so the change is visible
+            without waiting for the next mouse move. Anything else
+            raises `ValueError`.
+        """
         if val not in self.ALLOWED_SKY_COORDINATE_FORMATS:
             raise ValueError('Invalid value {} for sky_coordinate_format. '
                              'Valid values are: '
@@ -171,6 +206,13 @@ class CursorInfoMixin:
         """
         Update the readout for a cursor at ``(x, y)`` in data coordinates,
         using the data and WCS of the currently displayed image.
+
+        Parameters
+        ----------
+        x, y : float
+            Cursor position in data (pixel) coordinates, following the
+            `astropy.wcs.WCS.pixel_to_world` convention (pixel centers
+            at integer values, origin 0).
         """
         self._last_cursor_position = (x, y)
         self._refresh_cursor_text()
@@ -189,8 +231,6 @@ class CursorInfoMixin:
         data = info.data
         if isinstance(data, NDData):
             data = data.data
-        elif data is not None:
-            data = np.asarray(data)
 
         self._cursor_readout.value = READOUT_TEMPLATE.format(
             format_cursor_text(x, y, data=data, wcs=info.wcs,
--- a/astrowidgets/cursor_info.py
+++ b/astrowidgets/cursor_info.py
@@ -1,0 +1,197 @@
+"""
+Shared cursor-information readout for astrowidgets backends.
+
+Backends compose their image widget with an HTML readout that shows the
+cursor position (X/Y and, when a WCS is available, RA/Dec) and the image
+value under the cursor. The formatting and widget plumbing live here so
+that every backend behaves the same; each backend only forwards its
+native mouse-move events to `CursorInfoMixin._update_cursor_text`.
+"""
+import numpy as np
+
+from astropy import units as u
+from astropy.nddata import NDData
+
+import ipywidgets as ipw
+
+__all__ = ['format_cursor_text', 'CursorInfoMixin', 'READOUT_TEMPLATE',
+           'ALLOWED_CURSOR_LOCATIONS', 'ALLOWED_SKY_COORDINATE_FORMATS']
+
+ALLOWED_CURSOR_LOCATIONS = ('top', 'bottom', None)
+ALLOWED_SKY_COORDINATE_FORMATS = ('degrees', 'sexagesimal')
+
+# The readout is rendered in a <pre> so that the fixed-width padding in
+# format_cursor_text survives HTML whitespace collapsing and the digits
+# line up, keeping the line from jittering as the cursor moves.
+READOUT_TEMPLATE = '<pre style="margin: 0">{}</pre>'
+
+
+def format_cursor_text(x, y, data=None, wcs=None, sky_format='degrees'):
+    """
+    Format the cursor readout for a position in an image.
+
+    Parameters
+    ----------
+    x, y : float
+        Cursor position in data (pixel) coordinates, with pixel centers
+        at integer values (origin 0), i.e. the convention of
+        `astropy.wcs.WCS.pixel_to_world`.
+    data : array-like or None
+        Image data, indexed as ``data[y, x]``. When ``None``, or when the
+        cursor is outside the array, the value reads ``N/A``.
+    wcs : `astropy.wcs.WCS` or None
+        WCS used for the RA/Dec segment; omitted when ``None``. The sky
+        position is always converted to ICRS, whatever the native frame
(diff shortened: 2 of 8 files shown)


CHANGES.rst
```diff
@@ -3,9 +3,28 @@
 
 TBD -- First stable release.
 
+New Features
+------------
+
+- The cursor-information readout (X/Y, RA/Dec, pixel value) is now shared
+  between the ginga and bqplot backends via
+  ``astrowidgets.cursor_info.CursorInfoMixin``. The ``cursor`` property
```
mwcraig: We don't need the whole history of this in the change log entry.

astrowidgets/cursor_info.py
```diff
@@ -0,0 +1,197 @@
...
+                dec = sky.dec.to_string(unit=u.degree, sep=':',
+                                        precision=2, alwayssign=True,
+                                        pad=True)
+            else:
+                # Fixed width (RA 0-360, Dec always signed) so the line
+                # does not jitter as the cursor moves.
+                ra = f'{sky.ra.deg:<8.4f}'
+                dec = f'{sky.dec.deg:<+8.4f}'
+            segments.append(f'RA: {ra} Dec: {dec} (ICRS)')
+        except Exception:
```
mwcraig, line 98: Can this be a more specific exception?

astrowidgets/cursor_info.py
```diff
@@ -0,0 +1,197 @@
...
+    ALLOWED_SKY_COORDINATE_FORMATS = ALLOWED_SKY_COORDINATE_FORMATS
+
+    def _init_cursor_info(self, image_widget):
+        """
+        Create the readout widget and lay it out below ``image_widget``.
+        """
+        self._cursor_readout = ipw.HTML('Coordinates show up here')
+        self._sky_coordinate_format = 'degrees'
+        self._last_cursor_position = None
+        self.children = [image_widget, self._cursor_readout]
```
mwcraig: I do not like burying the setting of children in here-it makes it less visible to developers and means that it can only be used by ipywidget based

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,258 @@
+import numpy as np
+import pytest
+
+from astropy import units as u
+from astropy.wcs import WCS
+
+from astrowidgets.cursor_info import format_cursor_text, READOUT_TEMPLATE
+
+
+def _make_wcs():
```
mwcraig: Isn't there a make_wcs somewhere else in the tests?

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,258 @@
...
+    w = WCS(naxis=2)
+    w.wcs.crpix = [-234.75, 8.3393]
+    w.wcs.cdelt = np.array([-0.066667, 0.066667])
+    w.wcs.crval = [0, -90]
+    w.wcs.ctype = ["RA---AIR", "DEC--AIR"]
+    w.wcs.set_pv([(2, 1, 45.0)])
+    return w
+
+
+def _make_data():
```
mwcraig: Why aren't this and the WCS function fixtures?

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,258 @@
...
+    widget.sky_coordinate_format = 'sexagesimal'
+    assert widget.sky_coordinate_format == 'sexagesimal'
+
+    with pytest.raises(ValueError, match='sky_coordinate_format'):
+        widget.sky_coordinate_format = 'radians'
+
+
+@pytest.mark.parametrize('backend', ['bqplot', 'ginga'])
+def test_mouse_move_updates_readout(backend):
+    from astropy.nddata import NDData
```
mwcraig: Why is this import inside the
Function?

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,258 @@
...
+    widget = _backend_widget(backend)
+    _simulate_mouse_move(widget, backend, 3.0, 2.0)
+    assert widget._cursor_readout.value == 'Coordinates show up here'
+
+
+@pytest.mark.parametrize('backend', ['bqplot', 'ginga'])
+def test_sky_format_change_updates_readout_immediately(backend):
+    # Regression test: changing sky_coordinate_format used to leave the
+    # readout stale until the next mouse move.
+    from astropy.nddata import NDData
```
mwcraig: Same question as above.

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,258 @@
...
+    widget.sky_coordinate_format = 'sexagesimal'
+    assert widget._cursor_readout.value == 'Coordinates show up here'
+
+
+def test_backends_produce_identical_readout():
+    # The payoff of the shared module: both backends must produce the
+    # same readout string for the same cursor position and image.
+    pytest.importorskip('bqplot')
+    pytest.importorskip('ginga')
+    from astropy.nddata import NDData
```
mwcraig: ?

astrowidgets/cursor_info.py
```diff
@@ -0,0 +1,195 @@
...
+    # the negative side do not alias onto pixel 0.
+    x_index = int(np.floor(x + 0.5))
+    y_index = int(np.floor(y + 0.5))
+
+    if (data is not None
+            and 0 <= y_index < data.shape[0]
+            and 0 <= x_index < data.shape[1]):
+        value = f'value: {data[y_index, x_index]:8.3f}'
+    else:
+        value = 'value: N/A'
```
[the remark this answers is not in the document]
mwcraig: Confirmed — the docstring says "array-like" but the function goes straight to `data.shape` and 2D indexing, so a list-of-lists raises `AttributeError`. The widgets themselves are unaffected because `CursorInfoMixin._refresh_cursor_text` coerces with `np.asarray` before calling, but the standalone formatter doesn't live up to its own docs.

The clean fix is to move the coercion into the formatter so it is genuinely pure and reusable (a no-op for ndarray input):

```python
if data is not None:
    data = np.asarray(data)
```

at the top of `format_cursor_text`, and drop the now-redundant coercion from the mixin. Happy to apply this on the branch.

— _Written by Claude at @mwcraig's direction._

CHANGES.rst
mwcraig: Agreed — the "which had been dropped from the bqplot backend" narration is git history, not release notes. Proposed trimmed entry:

```rst
- The cursor-information readout (X/Y, RA/Dec, pixel value) is now shared
  between the ginga and bqplot backends via
  ``astrowidgets.cursor_info.CursorInfoMixin``. Both backends have a
  ``cursor`` property (``'top'``, ``'bottom'``, or ``None``) and a new
  ``sky_coordinate_format`` property that selects decimal degrees (the
  default) or sexagesimal for the RA/Dec display. [#222]
```

One thing I'd keep in some form: the ginga readout's default changed from sexagesimal to decimal degrees, which is a user-visible behavior change — but it can be a single sentence ("The ginga readout now defaults to decimal degrees; set ``sky_coordinate_format = 'sexagesimal'`` for the previous display.") rather than the current storytelling. Happy to apply on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/cursor_info.py
mwcraig: I'd argue the broad catch is the right call here, for two reasons:

1. **The exception surface is open-ended.** The WCS comes from whatever the user loaded and is only duck-typed to the APE-14 interface — `astropy.wcs.WCS`, `gwcs`, or anything else with `pixel_to_world`. Even for plain astropy WCS the chain `pixel_to_world(...).icrs` can raise `ValueError`, `astropy.coordinates.ConvertError` (frame conversion), or projection-specific errors; gwcs and other implementations raise their own types outside their valid region. Any tuple we write down will miss some of them.

2. **Where it runs matters.** This executes inside the mouse-move comm callback on every cursor movement. An exception that escapes there doesn't produce a clean traceback for the user — it either vanishes or spams the log once per pixel of mouse travel. The intent of the catch is "any WCS failure degrades to a visible ``RA/Dec: WCS error`` readout instead of a broken/silent widget."

Note `except Exception` already lets `KeyboardInterrupt`/`SystemExit` through (they're `BaseException`), so it isn't the fully-indiscriminate `except:`.

If you'd still prefer it narrower, the best concrete tuple is probably `(ValueError, TypeError, astropy.coordinates.ConvertError)` — but I'd rather keep the broad catch and add a comment explaining why it's deliberate. Your call; happy to do either on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/cursor_info.py
mwcraig: Fair on the visibility point — a mixin init silently replacing `self.children` is easy to miss when reading a backend's `__init__`. The minimal fix: `_init_cursor_info` only creates the readout widget and returns it, and each backend assembles its own children explicitly:

```python
# in each backend __init__
readout = self._init_cursor_info()
self.children = [self._astro_im, readout]   # visible in the backend itself
```

One honest caveat on the "only usable by ipywidgets-based backends" half: that coupling goes deeper than this line. The mixin also creates an `ipywidgets.HTML` and the `cursor` setter drives `layout.visibility`/`layout.display`/`flex_flow` — the 'top'/'bottom' placement *is* flexbox. Moving the `children` assignment out improves visibility but doesn't make the mixin backend-agnostic.

If a non-ipywidgets backend is a real prospect, the fuller split would be: keep the pure state + formatting in `cursor_info` (sky format, last position, produce the text string) and have backends implement a tiny display hook (`_set_readout_text(str)` plus their own placement logic). That's more surgery, so I'd suggest doing the minimal fix now and the split only when a concrete non-ipywidgets backend shows up. Happy to implement either on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
mwcraig: Yes — this is a third copy. `test_widget_api_ginga.py` has an identical `_make_wcs()` at line 21, and the same WCS (same crpix/cdelt/crval/AIR projection) is the `wcs` fixture in `astro_image_display_api.api_test.ImageAPITest`. The api-test one is a fixture method on the base test class, so it isn't importable as a plain function — the practical dedup is a `wcs` fixture in a new `astrowidgets/tests/conftest.py`, with both this module and `test_widget_api_ginga.py` using it. Happy to do that on the branch (pairs with the fixture question below).

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
mwcraig: No good reason — plain helpers were a style artifact, and fixtures are the pytest idiom. Converting `_make_wcs`/`_make_data` to `wcs`/`data` fixtures in a shared `conftest.py` also resolves the duplication you flagged above (the same WCS already exists in `test_widget_api_ginga.py` and as a fixture in `astro_image_display_api.api_test`). The couple of tests that call `_make_wcs()` twice (once to load, once to compute the expected string) work equally well with a single fixture instance. Will do on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
mwcraig: No good reason. The deferred-import pattern in this module exists for `bqplot`/`ginga`, which are optional dependencies that must only be imported after `importorskip`. `NDData` is astropy — a hard dependency — so its import belongs at the top of the module. This same local import appears three times (here, line 210, line 236); I'll move a single `from astropy.nddata import NDData` to the module top and delete all three. Will do on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
mwcraig: Same answer: no reason — astropy is a hard dependency, only the bqplot/ginga imports need deferring. Covered by moving `NDData` to the module top (see reply at line 187).

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
mwcraig: Same as the two above — this is the third copy of the local `NDData` import; the unusual placement (after the `importorskip` calls) suggests it was cargo-culted from the optional-backend imports, but `NDData` needs no skip guard. Moving it to the module top removes all three.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,250 @@
...
+    good = wcs.pixel_to_world(x, y).icrs
+    swapped = wcs.pixel_to_world(y, x).icrs
+    assert good.ra.deg != swapped.ra.deg
+
+    text = format_cursor_text(x, y, data=data, wcs=wcs)
+    assert f'RA: {good.ra.deg:<8.4f}' in text
+    assert f'Dec: {good.dec.deg:<+8.4f}' in text
+
+
+def test_format_sexagesimal(wcs):
```
mwcraig, line 41: Add a comment like in the other tests

astrowidgets/tests/test_cursor_info.py
```diff
@@ -0,0 +1,250 @@
...
+    sky = wcs.pixel_to_world(x, y)
+    assert sky.frame.name == 'galactic'
+    icrs = sky.icrs
+
+    text = format_cursor_text(x, y, wcs=wcs)
+    assert f'RA: {icrs.ra.deg:<8.4f}' in text
+    assert f'Dec: {icrs.dec.deg:<+8.4f} (ICRS)' in text
+
+
+def test_format_invalid_sky_format():
```
mwcraig, line 74: Even though it is hsort and obvious say what it is testing

astrowidgets/cursor_info.py
```diff
@@ -0,0 +1,206 @@
...
+    @property
+    def cursor(self):
+        """
+        Show or hide cursor information (X, Y, RA/Dec, value).
+        Acceptable values are 'top', 'bottom', or ``None``.
+        """
+        return self._cursor_location
+
+    @cursor.setter
+    def cursor(self, val):
```
mwcraig, line 146: Docstring please

astrowidgets/cursor_info.py
```diff
@@ -0,0 +1,206 @@
...
+    @property
+    def sky_coordinate_format(self):
+        """
+        Format of the RA/Dec part of the cursor readout, either
+        ``'degrees'`` (decimal degrees, the default) or ``'sexagesimal'``.
+        """
+        return self._sky_coordinate_format
+
+    @sky_coordinate_format.setter
+    def sky_coordinate_format(self, val):
```
mwcraig, line 184: Docstring

astrowidgets/cursor_info.py
```diff
@@ -0,0 +1,206 @@
...
+                             'Valid values are: '
+                             '{}'.format(val,
+                                         self.ALLOWED_SKY_COORDINATE_FORMATS))
+        self._sky_coordinate_format = val
+        self._refresh_cursor_text()
+
+    def _update_cursor_text(self, x, y):
+        """
+        Update the readout for a cursor at ``(x, y)`` in data coordinates,
+        using the data and WCS of the currently displayed image.
```
mwcraig, line 208: Add parameteres sesction to docstring

astrowidgets/tests/test_cursor_info.py
mwcraig: Agreed — its neighbors all say what they pin down and this one doesn't. Proposed:

```python
def test_format_sexagesimal(wcs):
    # The sexagesimal readout must match astropy's own to_string
    # rendering: RA as HH:MM:SS.ss (hours, zero-padded), Dec as
    # +/-DD:MM:SS.ss (always signed).
```

Will add on the branch.

— _Written by Claude at @mwcraig's direction._

example_notebooks/bqplot_widget.ipynb
```diff
@@ -721,6 +712,13 @@
     "print(\"Interactive marking is a bqplot-specific feature not in the API\")\n",
     "print(\"Use catalog-based marking instead with load_catalog()\")"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
```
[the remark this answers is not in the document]
mwcraig: Confirmed — the committed notebook's last cell is an empty code cell (`"source": []`). It's an artifact of editing in Jupyter, which appends a fresh cell after running the last one. Will remove it on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/tests/test_cursor_info.py
mwcraig: Agreed. Proposed:

```python
def test_format_invalid_sky_format():
    # An unrecognized sky_format must raise rather than silently fall
    # back to a default, and the error must name the bad argument.
```

Will add on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/cursor_info.py
mwcraig: Will do — the setter has enough going on (visibility, display, flex-direction flip) that a source reader benefits. One caveat so it's a conscious choice: Python exposes a property's documentation through the *getter* — `help(ImageWidget.cursor)` and Sphinx render the getter's docstring, and a setter docstring is invisible there. So the user-facing contract (allowed values, what they mean) stays on the getter, and the setter docstring documents the mechanics:

```python
@cursor.setter
def cursor(self, val):
    """
    Set the readout location: hide the readout for ``None``, or show
    it and flip the box's flex direction so it renders above
    (``'top'``) or below (``'bottom'``) the image. Re-renders the
    readout at the last cursor position. Raises ``ValueError`` for
    anything else.
    """
```

Happy to apply on the branch (same for ``sky_coordinate_format`` below).

— _Written by Claude at @mwcraig's direction._

astrowidgets/cursor_info.py
mwcraig: Will do, with the same caveat as the ``cursor`` setter above (setter docstrings don't surface in ``help()``/Sphinx, so the allowed values stay documented on the getter). Proposed:

```python
@sky_coordinate_format.setter
def sky_coordinate_format(self, val):
    """
    Validate the format against `ALLOWED_SKY_COORDINATE_FORMATS` and
    immediately re-render the readout at the last cursor position, so
    the change is visible without waiting for the next mouse move.
    """
```

— _Written by Claude at @mwcraig's direction._

astrowidgets/cursor_info.py
mwcraig: Agreed — numpydoc it is:

```python
Parameters
----------
x, y : float
    Cursor position in data (pixel) coordinates, following the
    `astropy.wcs.WCS.pixel_to_world` convention (pixel centers at
    integer values, origin 0).
```

That matches the wording already used for ``format_cursor_text``'s docstring, so the convention is stated identically in both places. Will add on the branch.

— _Written by Claude at @mwcraig's direction._

astrowidgets/cursor_info.py
mwcraig: Do a numpy-style docstring (i.e. add parameters)

astrowidgets/cursor_info.py
mwcraig: numpy-style with a parameters section

PR #232, opened 2026-08-23
Reduce redundant CI matrix jobs

### LINKED ISSUE #227: Reduce CI redundancy now that pixi carries an OS matrix
With the CI rearranged around pixi, each push/PR currently runs two full-suite OS matrices against release dependencies: `tests` (tox, 3 OSes) and `pixi` (3 OSes), plus `pixi-downstream` (3 OSes) which only pip-installs the package and runs `import astrowidgets`.

The tox and pixi jobs are complementary in kind, not duplicates — worth preserving both *paths*, but not both matrices:

| | tox `tests` | `pixi` |
|---|---|---|
| Dependencies | latest PyPI releases within `pyproject.toml` ranges (drifts; catches new-release breakage) | exact versions from `pixi.lock` via conda-forge (reproducible; catches lockfile/code breakage only) |
| Install mode | built sdist→wheel, installed into a clean venv, run from a separate directory (catches packaging bugs, e.g. files missing from the wheel) | editable install of the source tree (can't catch packaging bugs) |
| Scope | package tests + `docs/` | package tests only (plain `pytest` doesn't collect `docs/`) |

Proposed consolidation:

- [ ] Shrink tox `tests` to ubuntu-only, keeping it as the "PyPI-latest + real packaging + docs" job; let `pixi` carry the 3-OS matrix (OS-specific issues are mostly binary-dep issues, which conda-forge exercises well).
- [ ] Shrink `pixi-downstream` to ubuntu-only or drop it — its unique value is checking that astrowidgets-as-a-PyPI-dep solves inside a conda-forge pixi user env, which doesn't need three platforms for an import-only smoke test.
- [ ] Drop the `pixi` job's "Verify install" step (`pixi run python -c "import astrowidgets"`) — redundant with the test run that follows.
- [ ] Optionally: add `docs/` collection to the pixi `test` task so pixi jobs match tox's scope, and shrink the manual `rc_testing` workflow to ubuntu-only for consistency.

Net effect: ~5 fewer jobs per PR with essentially no coverage loss. Keep `devtests` (only job against astropy dev + ginga master; it caught ejeschke/ginga#1148) and pre-commit.ci (now the only lint check) as-is.

*Drafted by Claude Code on behalf of Matt Craig (@mwcraig).*

sapunyangkut:
## Summary

- keep the tox packaging and documentation test path on Ubuntu
- retain the three-platform pixi matrix for OS coverage
- run the downstream pixi install smoke test on Ubuntu and remove its now-unneeded platform branch
- remove the redundant import check from the pixi job, since the test step imports the package

This reduces the workflow from 10 runner jobs to 6 while preserving each distinct test path.

## Validation

- `actionlint 1.7.12 .github/workflows/ci_workflows.yml`
- parsed the workflow and checked the expected runner/matrix structure with PyYAML

Closes #227

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -17,11 +17,7 @@ permissions:
 
 jobs:
   tests:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: true
-      matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+    runs-on: ubuntu-latest
     steps:
     - name: Checkout code
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
@@ -62,16 +58,11 @@ jobs:
         run-install: false
     - name: Set up pixi environment
       run: pixi install
-    - name: Verify install
-      run: pixi run python -c "import astrowidgets"
     - name: Run tests (pixi)
       run: pixi run test
 
   pixi-downstream:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ubuntu-latest
     steps:
     - name: Checkout code
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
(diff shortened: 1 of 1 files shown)


mwcraig (approved):
Thanks @sapunyangkut!

